PR #35, opened 2024-02-21
New pp tools

ceblanton:
fre pp validate

fre pp install

fre pp run

fre pp status


ceblanton:
Hi @bcc2761 the python import and entrypoint tests worked, and I'm going to merge this in.

Issue #9, opened 2023-12-13
fre pp configure needs update to use reworked pp.yaml


ceblanton:
Currently, `fre pp configure -y pp.yaml` accepts a yaml file and writes 3 files in the current directory:
- rose-suite.conf
- app/regrid-xy/rose-app.conf
- app/remap-pp-components/rose-app.conf

The current pp.yaml looks like this (https://github.com/NOAA-GFDL/fre-cli/blob/main/fre/frepp/edits-template.yaml) and has an associated schema (https://github.com/NOAA-GFDL/fre-cli/blob/main/fre/frepp/schema.json).

Chris has reworked that pp.yaml structure to be more Bronx XML-like. Still up for debate / tweaks, though.

```
refineDiags:
  - script: "/path/to/abs_path/script"
  - script: "path/to/script/in/repo"

preAnalysis:
  - script: "/path/to/preanalysis/script"

settings:
  history_segment:
  start: 1980
  stop:
  site: "ppan"
  fre_analyis_home:
  pp_chunk_a:
  pp_chunk_b:
  grid_spec: "/path/to/tarfile"

switches:
  clean_work: true
  do_mdtf: false
  do_statics: true
  do_timeavgs: true
  do_refinediag: false
  do_atmos_plevel_masking: true
  do_preanalysis: false

directories:
  history: 
  pp:
  ptmp:
  history_refined:
  analysis:

metadata:
  experiment:
  platform:
  target:

components:
  - type: "atmos_cmip"
    source: atmos_month_cmip
    sourceGrid: atmos-cubedsphere
    xyInterp: 288,180
    interpMethod: conserve_order2
    timeSeries:
      - freq: 3hr
        source: atmos_8xdaily_cmip
        chunkLength: 1yr
      - freq: daily
        source: atmos_daily_cmip
        chunkLength: 1yr
        variables: huss tasmin tasmax tas pr psl sfcWind hurs uas vas zg500 ua10 ua850 va850 ua200
  - type: aerosol_cmip
    xyInterp: 288,180
    source: aerosol_month_cmip
    sourceGrid: atmos-cubedsphere
    interpMethod: conserve_order1
    timeSeries:
      - freq: "monthly"
        chunkLength: 1yr
      - freq: monthly
        chunkLength: 1yr
        source: atmos_month_refined
```

So we'd like `fre pp configure` to use this reworked pp.yaml instead.

PR #37, opened 2024-02-21
9.configure script update

singhd789:


singhd789:
Note: the configUpdates folder is not needed and will be taken out. The reason it is still in there in because of something Ian and I were looking at in our refactoring task, but it is solely for my development purposes.

fre/frepp/configUpdates/pp.yaml
```diff
@@ -0,0 +1,225 @@
...
+configuration_paths: 
+  rose-suite: "/home/Dana.Singh/pp/ppyaml/fre-cli/fre/frepp/configUpdates/rose-suite-am5_c96L33_amip.conf" 
+  rose-remap: "/home/Dana.Singh/pp/ppyaml/fre-cli/fre/frepp/configUpdates/remap-rose-app.conf"
+  rose-regrid: "/home/Dana.Singh/pp/ppyaml/fre-cli/fre/frepp/configUpdates/regrid-rose-app.conf"
+
+analysisLen: &ANA_AMIP_LEN "10yr"
+chunk96: &PP_AMIP_CHUNK96 "1yr"
+analysisStart: &ANA_AMIP_START "1980"
+analysisStop: &ANA_AMIP_STOP "1988"
+defxyInterp: &defaultxyInterp "180,288"
```
ceblanton, line 11: Neat... these are like FRE properties, user variables that can be reused later?

I see the value set there, e.g. `10yr`, but is the variable `analysisLen` or `ANA_AMIP_LEN`?

fre/frepp/configUpdates/pp.yaml
```diff
@@ -0,0 +1,225 @@
...
+
+rose-suite:
+  refineDiag:
+    datastager_script: "$(includeDir)/refineDiag/refineDiag_data_stager_globalAve.csh"
+    atmoscmip6_script: "$(includeDir)/refineDiag/refineDiag_atmos_cmip6.csh"
```
ceblanton, line 16: Is "atmoscmip6_script" part of the yaml schema or a user setting?

fre/frepp/configUpdates/pp.yaml
```diff
@@ -0,0 +1,225 @@
...
+    history_refined: "/archive/$USER/canopy/am5/c96L33_amip/history_refineDiag"
+    analysis: #"/nbhome/$USER/canopy/analysis/am5_c96L33_amip"
+
+  metadata:
+    experiment: "am5_c96L33_amip"
+    platform: "gfdl.ncrc5-intel22-classic"
+    target: "prod-openmp"
+
+#regrid-xy and remap-pp-components information
+regrid-remap-info:
```
ceblanton, line 53: Yes! The `components` part down is perfect as is, clearly analogous to a Bronx XML and I think users will be able to understand it.

Can the `regrid-remap-info` header be removed?

fre/frepp/configUpdates/schema-pp.json
```diff
@@ -0,0 +1,137 @@
...
+    "configuration_paths": { 
+      "type": "object",
+      "properties": {
+        "rose-suite": {"type": "string"},
+        "rose-remap": {"type": "string"},
+        "rose-regrid": {"type": "string"}
+      }
+    },
+    "analysisLen": {"type": "string"},
+    "chunk96": {"type": "string"},
```
ceblanton, line 15: The name `chunk96` is part of the schema, here?

ceblanton (approved):
Dana, thank you so much for this. I can't follow all of it, and I think some of the pp.yaml could be simplified, but if mostly works now, I'm inclined to think we should merge it and begin to use and test it more. We can always adjust as we go.

fre/frepp/configUpdates/pp.yaml
singhd789: Yes, it can. I put that in there for more organization but I'm sure it would work with just `components` as the header here.

fre/frepp/configUpdates/pp.yaml
singhd789: Oh! This was originally following the draft you had started in the issue (https://github.com/NOAA-GFDL/fre-cli/issues/9). There were 2 refineDiag scripts but both were called `script` so I found that the latter was the only one found. Changing the names from `script` just let's us have access to both scripts.

fre/frepp/configUpdates/pp.yaml
singhd789: Yes! They're used later in the components section. It was slightly confusing, the variable is defined as `analysisLen` in the yaml, but the value actually reused later, that saves the value the user sets, is the one with the `&`

fre/frepp/configUpdates/schema-pp.json
singhd789: Oo, I'll check this.

singhd789: Yeah, the `chunk96` was part of those user variables that could be reused later. This name could change

singhd789:
Thanks for taking a look through this Chris! 

I attempted to test this before, to see if the configurations were actually good to run a workflow, but there were some error messages in the `rose macro --validate` step. I know this has it's own issues, but it brought to light that I may have forgot some variables or potentially set some incorrectly, like `outputGridType` or `inputRealm` in rose-apps. I'm in the process of looking through these.

fre/frepp/configUpdates/schema-pp.json
ceblanton: User variable names shouldn't be in the pp.yaml schema, should they? Can users name their variables whatever they like?

singhd789: Well in the pp.yaml, in order for it to be validated, should be defined as something. I can change it to something like `pp_amip_chunk` or something instead of `chunk96`. The actual variable that can be reused throughout the pp.yaml, i.e. ` &PP_AMIP_CHUNK96`, can be called anything the user wants though, as long as it's consistent through the yaml.

ceblanton: Sorry, I'm still confused. is the variable called `pp_amip_chunk` or `PP_AMIP_CHUNK96`?

singhd789: You're good! That wasn't a great explanation on my part at all. I was looking for this link to help. 

https://docs.geoserver.org/2.21.x/en/user/styling/ysld/reference/variables.html

With reusable yaml variables, it has this structure:
`define: &variable <value>`
Where: `define = pp_amip_chunk`
              `$variable = &PP_AMIP_CHUNK96` (which was what the variable was called in the XML, and the variable to be reused in the yaml later on in the components section)
              `<value> = "1yr"`

ceblanton: aha, thanks for that. I see. I agree that's just what we want.

Can't we have just one variable name, `PP_AMIP_CHUNK96` that is user set? What does the lower case `pp_amip_chunk` in?

This is the stock example:
```
define: &variable <value>
```

Here's our example:
```
define: %PP_AMIP_CHUNK96 '1yr'
```

And then PP_AMIP_CHUNK96 can be used in the yaml below as:

```
*PP_AMIP_CHUNK96
```

fre/frepp/configUpdates/pp.yaml
singhd789: Quick clarifying question here, in the `rose-suite.conf`, the variable `REFINEDIAG_SCRIPTS` is sometimes defined (if do_refineDiag is True). Would that variable need to point to both of these scripts?

singhd789: I see that in past rose-suites, the data_stager script was used for the `preanalysis_script` and the atmoscmip  script, used for the refineDiag_scripts variables. Should we just keep it like that and add those to the directories section with their respective variables (preanalysis and refinediag ...)?

PR #42, opened 2024-03-01
adding notification for which directory cloned into

cwhitlock-NOAA:


cwhitlock-NOAA:
Yes, it's just the print statement. It's a hello world for understanding the merge process.

bcc2761:
@cwhitlock-NOAA Sorry I didn't get back to you on Friday; I actually worked on a fix for this as soon as I saw your issue submitted, but I still want to get your merge in. Would you mind making a new commit with this change?

https://github.com/bcc2761/fre-cli/commit/e48e227695907e464317878fbab29bc87f749f45

Click prefers `click.echo` over `print` (reference: https://click.palletsprojects.com/en/8.1.x/quickstart/#echoing), and I thought that moving this notification to the end would be better than before the cloning occurs. Let me know if you have any thoughts

bcc2761:
Good point. I didn't account for failed clones that might confuse the user. Yeah, we can add it before the clone occurs then

PR #43, opened 2024-03-01
Pressure masking netcdf modifying tool

ceblanton:


ceblanton:
This is a neat self-contained tool that is useful to for FRE workflows, but it needs more documentation, tests, and it's not working perfectly yet in comparison to the Bronx refinediag script.

bcc2761:
@ceblanton I'm going to go in and change the argparse to click while I review this

fre/freapp/maskAtmosPlevel.py
chanwilson, line 1: This is not how one adds arguments to a sub-command using Click!

chanwilson (changes requested):
Click is the way arguments are handled, not with argparse!

fre/freapp/maskAtmosPlevel.py
chanwilson: oh i didn't see your earlier comment.  so, echo chamber.

chanwilson:
I'm also not overjoyed about the qty of changes required when adding sub-commands, I have a couple thoughts on how to make that better.... but not for this PR.

fre/freapp/maskAtmosPlevel.py
ceblanton: Even though the maskAtmosPlevel.py script uses argparse (pointlessly), it is still called through click.

You can `pip install .` this over a conda-installed `fre-cli` environment to show it.

ceblanton: Done (argparse is removed)

ceblanton:
This really needs a single pytest.

bcc2761 (approved):

PR #47, opened 2024-03-15
fre pp convert: Remove unneeded check for PATH.

ceblanton:
The current check is not portable and fails on gaea.


ceblanton:
@bcc2761 this is a tiny victimless change that I would normally sneak in without any review, ha.

bcc2761 (approved):

PR #51, opened 2024-03-18
Sphinx Documentation

Ciheim:
Setting up Sphinx


docs/.doctrees/FAQ.doctree
ceblanton, line 1: Is this binary file (and the next few) needed, or is it generated by sphinx as an intermediate file or such?

ceblanton (changes requested):
Looks good, except I don't like the binary files in git repos. Are they really needed?

docs/.doctrees/FAQ.doctree
Ciheim: Auto generated and not needed. Removing!

ceblanton (approved):
This looks great, thank you!

PR #58, opened 2024-03-20
57.update configure yaml

singhd789:
Add in `-e`, `-p`, and `-t` options


fre/frepp/ppyaml_test/pp.yaml
```diff
@@ -1,9 +1,9 @@
 ## User defined edits
+#User defined paths for where the rose-suite and rose-app configurations should be
 configuration_paths: 
-## DEBUG
-  rose-suite: "$PWD/rose-suite.conf"
-  rose-remap: "$PWD/remap-rose-app.conf"
-  rose-regrid: "$PWD/regrid-rose-app.conf"
+  rose-suite: "./rose-suite.conf"
+  rose-remap: "./remap-rose-app.conf"
+  rose-regrid: "./regrid-rose-app.conf"
```
ceblanton, line 6: Are these rose paths required or optional? I think it would be better to make them optional or omit them if possible.

ceblanton (approved):
Looks good. I think we should remove the rose mentions from the yaml but we can take that on in another issue.

fre/frepp/ppyaml_test/pp.yaml
singhd789: They are currently required to create the rose configs (but will go away when rose goes away). These are paths that the user has to define for where the configs have to go in the pp repo. 

I think in a previous mr for the pp yaml, you mentioned these could possible be click options instead so they're not in here. I thought that would be a lot of click options for the user to input (ex: -e, -p, -t, -y, -rosesuite, -regrid-rose, -remap-rose), but maybe it would be good to take them out of here

PR #60, opened 2024-03-20
changed setup configurations

bcc2761:
to get rid of "package will be ignored" message in workflow runs


bcc2761:
I thought this would fix it but it didn't... not the biggest problem but just something bothersome

bcc2761:
Fixed, but "clobber warnings" are back...

PR #65, opened 2024-03-27
56.fre pp configure updates

singhd789:


fre/frepp/configureScriptYAML.py
```diff
@@ -23,66 +24,46 @@ def validateYaml(file):
...
   p = platform
-  t = target 
+  t = target
+  yml = yamlfile
 
-  with open(y,'r') as f:
+  with open(yml,'r') as f:
     y=yaml.safe_load(f)
-    yml = validateYaml(y)
+    valyml = validateYaml(y)
```
ceblanton, line 35: `y` is no longer set.. should this be `validateYaml(yml)` instead?

ceblanton (changes requested):
Looks good. I think you renamed `y` to `yml` in one section that needs one more change, though.

fre/frepp/configureScriptYAML.py
singhd789: The yamlfile is being saved as `yml`, which is being opened as `f` and then `safe_loaded` as `y`. So the loaded yaml is still `y`.

singhd789: Before I had overwritten variables accidentally, using `y` multiple times I think

ceblanton: Gotcha!

ceblanton (approved):

PR #67, opened 2024-03-28
Refactoring Attempt #1

bcc2761:
First steps towards refactorization by changing the way module imports are done, and also removing some duplicate boilerplate code within project. Removed tool.py files like frepp.py, also renamed all folders to simpler more straightforward tool names and removed 'fre' preceding their names. Renamed groups within fre.py to account for this


bcc2761:
This is obviously causing a lot of merge conflicts so we can hold onto it for now

singhd789 (approved):

PR #80, opened 2024-04-19
 Update location and print statements for rose config files

singhd789:
- remove optional configurations: just keep default `rose-suite.conf`


ceblanton (approved):
Looks great!

PR #83, opened 2024-04-24
Refactoring & README 

bcc2761:
Refactoring Part 2, focused on integrating Click's LazyGroup class to reduce imports


bcc2761:
Todo: update README to follow this; going to implement after approved

fre/make/.fremake.py.swp
chanwilson, line 1: this is a vi temporary file, please remove from commit.

chanwilson (approved):
Looks great!  Please remove the .swp file that snuck in, otherwise it's good to go in.

fre/make/.fremake.py.swp
bcc2761: Ah yes I was confused as to why this was even included. Does it matter, since it's being deleted here? How do I go about removing a file from a commit?

chanwilson: If it's being removed by this commit, that's great.  I thought it was being added...

chanwilson (approved):

fre/make/.fremake.py.swp
ceblanton: we have a .gitignore file but it doesn't ignore the .swp files yet.

https://github.com/NOAA-GFDL/fre-cli/blob/main/.gitignore

ceblanton (approved):

ceblanton:
Agree with Chan. This is such a big improvement. Thank you!

PR #86, opened 2024-05-07
 Fix grid value name for remap and regrid rose apps (for now)

singhd789:
In the remap-pp-components rewrite, Dana realized the grid value needed to specify default or whatever grid was being used, in the rose-app.conf. This pr is fixing the way `grid` is configured

### DIFF
--- a/fre/pp/configureScriptYAML.py
+++ b/fre/pp/configureScriptYAML.py
@@ -132,9 +132,9 @@ def yamlInfo(yamlfile,experiment,platform,target):
                 if i.get("xyInterp") == None:
                   f.write(f"grid=native\n")
                 #in xyInterp exists, component can be regridded
-                elif i.get("xyInterp") != None and i.get("xyInterp") == y["define5"]:
+                elif i.get("xyInterp") != None and i.get("xyInterp") == y["defaultxyInterp"]:
                   f.write("grid=regrid-xy/default\n") 
-                elif i.get("xyInterp") != None and i.get("xyInterp") != y["define5"]:
+                elif i.get("xyInterp") != None and i.get("xyInterp") != y["defaultxyInterp"]:
                   gridLat=i.get("xyInterp").split(",")[0]
                   gridLon=i.get("xyInterp").split(",")[1]
                   f.write(f"grid=regrid-xy/{gridLat}_{gridLon}\n")
@@ -180,7 +180,7 @@ def yamlInfo(yamlfile,experiment,platform,target):
 
                   f.write("\n")
 
-                if compkey == "xyInterp" and compvalue == y["define5"]:
+                if compkey == "xyInterp" and compvalue == y["defaultxyInterp"]:
                   f.write(f"outputGridType=default\n")
                 elif compkey == "xyInterp": 
                   gridLat=compvalue.split(",")[0]
--- a/fre/pp/configureScriptYAML.py
+++ b/fre/pp/configureScriptYAML.py
@@ -132,10 +132,14 @@ def yamlInfo(yamlfile,experiment,platform,target):
                 if i.get("xyInterp") == None:
                   f.write(f"grid=native\n")
                 #in xyInterp exists, component can be regridded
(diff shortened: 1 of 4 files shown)


fre/pp/configureScriptYAML.py
```diff
@@ -132,10 +132,14 @@ def yamlInfo(yamlfile,experiment,platform,target):
                 if i.get("xyInterp") == None:
                   f.write(f"grid=native\n")
                 #in xyInterp exists, component can be regridded
-                elif i.get("xyInterp") != None:
-                  f.write("grid=regrid-xy\n") 
+                elif i.get("xyInterp") != None and i.get("xyInterp") == y["define5"]:
+                  f.write("grid=regrid-xy/default\n") 
+                elif i.get("xyInterp") != None and i.get("xyInterp") != y["define5"]:
```
ceblanton, line 137: Hmmm what's this `define5` stuff? It's needed, isn't it?

If it's needed, maybe looking for something more less specific would be reasonable, especially as we're dumping Rose app configs anyway.

Maybe checking for the regular expression `define[0-9][0-9]` or such.

singhd789: the `y[define5]` is specifically looking at/parsing the key defined at the top in the yamls:

https://github.com/NOAA-GFDL/fre-cli/blob/088ad363392b3bf187119d8970c22779d59aaed0/fre/pp/ppyaml_test/pp.yaml#L12

ceblanton: Thanks, I see- I have mixed feelings about the yaml user variable definitions as we have them. They're exactly what we want with built-in yaml feature, but setting a user variable `defaultxyInterp` or such is more self-documenting and extensible than `define5`.

e.g. couldn't a different equally valid pp.yaml have the user-defined setting as `define4` or define6`?

singhd789: Of course this is just for that example ppyaml I have in here

ceblanton: After looking more, I see why you did it this way, and I'm merely disagreeing with the `grid=regrid-xy/default` convention that I started!

The confusing thing (to me) is that the `configure-yaml` script is looking for the label of the user variable, when I think it should only be looking at the yaml `xyInterp` setting (regardless of what the user variable name is).

You have this totally sensible logic:

For each component,

1. if `xyInterp` is not set, then use `grid=native` for the remap file (and nothing for the regrid file)

2. If `xyInterp` is set to a default, then use `grid=regrid-xy/default` (and `outputGridType=default` in the regrid file)

3. If `xyInterp` is set to a normal value, then use `grid=regrid-xy/{gridLat}-{gridLon}` in the remap file and `outputGridType={gridLat}-{gridLon}` for the remap file.

I'm suggesting we omit the second case entirely. `xyInterp` will always resolve to a `lat_lon` label whether it is set in yaml normally or set through the yaml via user defined variable. So setting the grid type (output of regrid file, input for remap file) to `lat_lon` rather than `default` is totally identical. (But still confusing, I agree.)

Another way of phrasing this very slight quibble is that the user variables are going to be totally arbitrary and can't be relied on in configure-yaml.py. e.g. "my_variable" should be just as valid for a valid pp.yaml

```
define5: &defaultxyInterp "180,288"
xyInterp: *defaultxyInterp  
```

is the same as

```
defaultxyInterp: &defaultxyInterp "180,288"
xyInterp: *defaultxyInterp  
```

or even

```
my_variable: &defaultxyInterp "180,288"
xyInterp: *my_variable
```

ceblanton: e.g. the configure-xml.py script doesn't use the default trick, and only uses the method 3 you have above.

https://github.com/NOAA-GFDL/fre-cli/blob/main/fre/pp/configureScriptXML.py#L399-L403

https://github.com/NOAA-GFDL/fre-cli/blob/main/fre/pp/configureScriptXML.py#L451

ceblanton (approved):

PR #91, opened 2024-05-13
improvements/updates to configureScriptXML + environment.yml updated

ilaflott:
put `environment.yml` back in with updated requirements, in case one wants to build the environment "by-hand" instead of using a conda channel. `pylint` and `pytest` added in

most updates to `fre/pp/configureScriptXML.py` are suggestions from `pylint`.


bcc2761:
Looks good to me; we technically don't need to add something under the setup.py requirements if it's already in the meta.yaml Conda requirements. We had cylc in there in the past but removed it. Not sure if having it both conda installed and pip installed will interfere with anything; this is probably why I removed it but can't recall completely. Going to ask for a second pair of eyes to look over the lines I'm less familiar with

bcc2761 (commented):
Looks good to me; we technically don't need to add something under the setup.py requirements if it's already in the meta.yaml Conda requirements. We had cylc in there in the past but removed it. Not sure if having it both conda installed and pip installed will interfere with anything; this is probably why I removed it but can't recall completely. Going to ask for a second pair of eyes to look over the lines I'm less familiar with

environment.yml
```diff
@@ -0,0 +1,16 @@
...
+  - python
+  - click
+  - pyyaml
+  - pylint
+  - jsonschema
+  - noaa-gfdl::intakebuilder
+  - conda-forge::cylc-flow
+  - conda-forge::cylc-rose
+  - conda-forge::metomi-rose
+  - conda-forge::cmor
```
ceblanton, line 16: I don't disagree with an environment.yaml here but is it a duplicate of meta.yaml? We might forget to update this one. Is there something here that's not in the meta.yaml?

fre/pp/__init__.py
```diff
@@ -1,6 +1,6 @@
 from .checkoutScript import checkoutTemplate
-from .configureScriptYAML import yamlInfo
-from .configureScriptXML import convert
+from .configure_script_yaml import yamlInfo
+from .configure_script_xml import convert
```
ceblanton, line 3: Just curious, was this a cosmetic improvement or necessary? I prefer snakes or camels myself so agree with this and more.

fre/pp/configure_script_xml.py
```diff
@@ -174,67 +193,93 @@ def main(xml, platform, target, experiment, do_analysis, historydir, refinedir,
...
+                                               universal_newlines=True)
+        historydir = fetch_history_process.stdout.strip() + '/history'
+        logging.info(historydir)
 
-    gridSpec = frelist_xpath(xml, platform, target, experiment, do_analysis, historydir, refinedir, ppdir, do_refinediag, pp_start, pp_stop, validate, verbose, quiet, dual, 'input/dataFile[@label="gridSpec"]')
-    simTime = frelist_xpath(xml, platform, target, experiment, do_analysis, historydir, refinedir, ppdir, do_refinediag, pp_start, pp_stop, validate, verbose, quiet, dual, 'runtime/production/@simTime')
-    simUnits = frelist_xpath(xml, platform, target, experiment, do_analysis, historydir, refinedir, ppdir, do_refinediag, pp_start, pp_stop, validate, verbose, quiet, dual, 'runtime/production/@units')
+
+
+    # Q: should respond to do_refineDiag and refinedir args of this function?
```
ceblanton, line 213: It's not an answer, but in general we want to hide all refinediag stuff and hope no one misses it. Preserve the capability and discourage its use. :)

fre/pp/configure_script_xml.py
```diff
@@ -174,67 +193,93 @@ def main(xml, platform, target, experiment, do_analysis, historydir, refinedir,
...
+                                          shell=True,
+                                          check=True,
+                                          capture_output=True,
+                                          universal_newlines=True)
+        ppdir = fetch_pp_process.stdout.strip()
+        logging.info(ppdir)
+
+
+    # Q: shouldn't there be a CLI analysis dir arg while we're here?
+    # basically, this is borderline on the same level as the ppDir and historydir fields.
```
ceblanton, line 232: It's painful to have to retrieve them separately with separate frelist calls, is that what you're saying?

setup.py
```diff
@@ -11,7 +11,10 @@
     install_requires=[
         'click',
         'pyyaml',
+        'pylint', 
         'jsonschema',
+        'cylc-flow',
+        'cylc-rose',
```
ceblanton, line 17: @bcc2761 brought this up somewhere else. I think this is OK, but probably won't be used.

`pip install .` called from the meta.yaml won't install prerequisites (it sort of assumes the prerequisites can be installed through conda), but interactive fre-cli developers who `pip install .` outside of conda might take advantage of this

ceblanton (approved):

fre/pp/configure_script_xml.py
ilaflott: reasonable!

setup.py
ilaflott: i am probably that sole fre-cli dev... this is clearly not generally needed.

fre/pp/configure_script_xml.py
ilaflott: more like, since `ANALYSIS_DIR` can be set by hand in a config file and everything works out just fine, might as well make it an input arg IMO. 

sometimes, being able to make I/O directories explicit is nice

fre/pp/__init__.py
ilaflott: "cosmetic", implying, "please, pylint, stop yelling at me".

but, as a matter of taste, and convention, mixed/camel case in python is reserved for certain contexts, i.e. classes. perhaps these blocks of code were slated for promotion to a class, @bcc2761 ?

environment.yml
ilaflott: essentially, yes.

my thinking here is that `environment.yml` builds a bare environment into which the `fre-cli` can be pip-installed as the sole instance of `fre`. That being said, from in-person conversations with @bcc2761 , this doesn't seem to gain us anything over `pip install`'ing `fre-cli` on top of the conda package one and explicitly invoking `fre` using the full path to the conda environment's executable

fre/pp/__init__.py
bcc2761: @ilaflott @ceblanton Nothing significant, just that my Java brain was wired to use CamelCase for some reason and didn't realize Python doesn't usually do the same

setup.py
bcc2761: > @bcc2761 brought this up somewhere else. I think this is OK, but probably won't be used.
> 
> `pip install .` called from the meta.yaml won't install prerequisites (it sort of assumes the prerequisites can be installed through conda), but interactive fre-cli developers who `pip install .` outside of conda might take advantage of this

Correct, this is why I had trouble integrating my 'gfdlfremake' pypi package.

PR #94, opened 2024-05-21
PP configure-yaml updates

ceblanton:
Minor cleanup:
* use metomi.rose for writing rose config files
* retrieve pathnames from command line arguments instead of pp.yaml

Also simplify some of the pp.yaml component definitions, by removing timeseries and timeaverage tags that are not absolutely needed in canopy.

### DIFF
--- a/fre/pp/configure_script_yaml.py
+++ b/fre/pp/configure_script_yaml.py
@@ -99,7 +99,7 @@ def yamlInfo(yamlfile,experiment,platform,target):
         # if xyInterp exists, component can be regridded
         else:
           interp_split = i.get('xyInterp').split(',')
-          rose_remap.set(keys=[f'{comp}', 'grid'], value=f'regrid-xy/{interp_method[1]}_{interp_method[0]}.{interp_method}')
+          rose_remap.set(keys=[f'{comp}', 'grid'], value=f'regrid-xy/{interp_split[0]}_{interp_split[1]}.{interp_method}')
 
         # set regrid items
         if i.get("xyInterp") != None:
@@ -110,7 +110,7 @@ def yamlInfo(yamlfile,experiment,platform,target):
           interp_split = i.get('xyInterp').split(',')
           rose_regrid.set(keys=[f'{comp}', 'outputGridLon'], value=f'{interp_split[1]}')
           rose_regrid.set(keys=[f'{comp}', 'outputGridLat'], value=f'{interp_split[0]}')
-          rose_regrid.set(keys=[f'{comp}', 'outputGridType'], value=f'{interp_method[1]}_{interp_method[0]}.{interp_method}')
+          rose_regrid.set(keys=[f'{comp}', 'outputGridType'], value=f'{interp_split[0]}_{interp_split[1]}.{interp_method}')
 
   # write rose configs
   print("Writing output files...")
--- a/fre/pp/configure_script_yaml.py
+++ b/fre/pp/configure_script_yaml.py
@@ -6,6 +6,7 @@ import click
 from jsonschema import validate, ValidationError, SchemaError
 import json
 import shutil
+import metomi.rose.config
 
 ######VALIDATE#####
 package_dir = os.path.dirname(os.path.abspath(__file__))
@@ -30,6 +31,28 @@ def yamlInfo(yamlfile,experiment,platform,target):
   t = target
   yml = yamlfile
 
+  # initialize rose suite config
+  rose_suite = metomi.rose.config.ConfigNode()
+  rose_suite.set(keys=['template variables', 'SITE'],              value='"ppan"')
+  rose_suite.set(keys=['template variables', 'CLEAN_WORK'],        value='True')
+  rose_suite.set(keys=['template variables', 'PTMP_DIR'],          value='"/xtmp/$USER/ptmp"')
+  rose_suite.set(keys=['template variables', 'DO_STATICS'],        value='True')
+  rose_suite.set(keys=['template variables', 'DO_TIMEAVGS'],       value='True')
+  rose_suite.set(keys=['template variables', 'DO_ATMOS_PLEVEL_MASKING'], value='True')
+
+  # initialize rose regrid config
+  rose_regrid = metomi.rose.config.ConfigNode()
+  rose_regrid.set(keys=['command', 'default'], value='regrid-xy')
+
+  # initialize rose remap config
+  rose_remap = metomi.rose.config.ConfigNode()
+  rose_remap.set(keys=['command', 'default'], value='remap-pp-components')
+
+  # set some rose suite vars
+  rose_suite.set(keys=['template variables', 'EXPERIMENT'], value=f'"{e}"')
+  rose_suite.set(keys=['template variables', 'PLATFORM'], value=f'"{p}"')
+  rose_suite.set(keys=['template variables', 'TARGET'], value=f'"{t}"')
+
   with open(yml,'r') as f:
     y=yaml.safe_load(f)
     valyml = validateYaml(y)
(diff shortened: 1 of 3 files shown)


fre/pp/configure_script_yaml.py
```diff
@@ -30,165 +31,94 @@ def yamlInfo(yamlfile,experiment,platform,target):
...
+  rose_suite.set(keys=['template variables', 'SITE'],              value='"ppan"')
+  rose_suite.set(keys=['template variables', 'CLEAN_WORK'],        value='True')
+  rose_suite.set(keys=['template variables', 'PTMP_DIR'],          value='"/xtmp/$USER/ptmp"')
+  rose_suite.set(keys=['template variables', 'DO_STATICS'],        value='True')
+  rose_suite.set(keys=['template variables', 'DO_TIMEAVGS'],       value='True')
+  rose_suite.set(keys=['template variables', 'DO_ATMOS_PLEVEL_MASKING'], value='True')
+  # disagreeable; these should be optional
+  rose_suite.set(keys=['template variables', 'DO_ANALYSIS_ONLY'],  value='False')
+  rose_suite.set(keys=['template variables', 'DO_MDTF'],  value='False')
+  rose_suite.set(keys=['template variables', 'PP_DEFAULT_XYINTERP'],  value='0,0')
```
ceblanton, line 45: I'm not wild about setting these defaults that really shouldn't be required, but they're currently required.

ceblanton:
@singhd789 and @Ciheim is this update OK to go in? I'm afraid it will affect your ppp configurations but it's worth it. (I was shamelessly going to click review/merge it in but the person who submitted the PR is not allowed to review by our repo rules..)

fre/pp/ppyaml_test/pp.yaml
```diff
@@ -1,225 +1,79 @@
...
-      - freq: "monthly"
-        chunkLength: *PP_AMIP_CHUNK96
-
-tmpdir:
-  tmpdirpath:
-
-install-exp-script:
-    - path: 
-    - install-option:
-        install:
```
singhd789, line 225: I wasn't sure if we still wanted these or not. Currently, they are not being used in the configure yaml script, but they are useful for running on `generic` sites (non-container use: stage-history used tmp path and the `install-option`  adds on a `symlink-dirs ...` to the cylc install command, if the user wanted the cylc-run directory to be put in a location with more space). 

Just a clarification, should we keep these for a non-container use on other sites? (and then I guess add onto the configure yaml script eventually)

fre/pp/ppyaml_test/pp.yaml
```diff
@@ -1,225 +1,79 @@
-## User defined edits
-# Paths for where the rose-suite and rose-app configurations are in checked out pp repo
```
singhd789, line 2: We need to also make sure that changes here are reflected in the schema.json so the pp.yaml can be validated correctly

fre/pp/configure_script_yaml.py
```diff
@@ -30,165 +31,94 @@ def yamlInfo(yamlfile,experiment,platform,target):
...
+        # set regrid items
+        if i.get("xyInterp") != None:
+          rose_regrid.set(keys=[f'{comp}', 'sources'], value=f'{sources}')
+          rose_regrid.set(keys=[f'{comp}', 'inputRealm'], value=f'{i.get("inputRealm")}')
+          rose_regrid.set(keys=[f'{comp}', 'inputGrid'], value=f'{i.get("sourceGrid")}')
+          rose_regrid.set(keys=[f'{comp}', 'interpMethod'], value=f'{interp_method}')
+          interp_split = i.get('xyInterp').split(',')
+          rose_regrid.set(keys=[f'{comp}', 'outputGridLon'], value=f'{interp_split[1]}')
+          rose_regrid.set(keys=[f'{comp}', 'outputGridLat'], value=f'{interp_split[0]}')
+          rose_regrid.set(keys=[f'{comp}', 'outputGridType'], value=f'{interp_split[0]}_{interp_split[1]}.{interp_method}')
```
singhd789, line 100: Just double checking, so for the outputGridType, the output would look like this:
```outputGridType=180_288.conserve_order2```

Instead of previously, like this, for example:
```outputGridType=180_288```

Similar to `grid` comment

File: fre/pp/configure_script_yaml.py [hunk 5]
```diff
@@ -30,165 +31,94 @@ def yamlInfo(yamlfile,experiment,platform,target):
...
+
+        # set remap items
+        rose_remap.set(keys=[f'{comp}', 'sources'], value=f'{sources}')
+        # if xyInterp doesnt exist, grid is native
+        if i.get("xyInterp") == None:
+          rose_remap.set(keys=[f'{comp}', 'grid'], value='native')
+        # if xyInterp exists, component can be regridded
+        else:
+          interp_split = i.get('xyInterp').split(',')
+          rose_remap.set(keys=[f'{comp}', 'grid'], value=f'regrid-xy/{interp_split[0]}_{interp_split[1]}.{interp_method}')
```
singhd789, line 89: Just checking for clarification and for if this is what we wanted. For grid, the output would look like this:
```grid=regrid-xy/180_288.conserve_order2```

Instead of previously, like this, for example:
```grid=regrid-xy/180_288```

**Comment by singhd789** [comment]
> @singhd789 and @Ciheim is this update OK to go in? I'm afraid it will affect your ppp configurations but it's worth it. (I was shamelessly going to click review/merge it in but the person who submitted the PR is not allowed to review by our repo rules..)

Hey Chris, sorry for the delay. I just added some comments

File: fre/pp/configure_script_yaml.py (hunk 5 above)
Comment (ceblanton) [reply fre/pp/configure_script_yaml.py:89]: Yes, I think so. the remap files are particular to the interpolation method

File: fre/pp/configure_script_yaml.py (hunk 4 above)
Comment (ceblanton) [reply fre/pp/configure_script_yaml.py:100]: Yep. those have to match until we can retire the rose app configs.. and then have those two apps use the pp.yaml directly.

File: fre/pp/ppyaml_test/pp.yaml (hunk 3 above)
Comment (ceblanton) [reply fre/pp/ppyaml_test/pp.yaml:2]: Thank you, I hadn't even thought of that. The good news is I see the "yaml valid" message each time during my testing so it may be ok this case?

File: fre/pp/ppyaml_test/pp.yaml (hunk 2 above)
Comment (ceblanton) [reply fre/pp/ppyaml_test/pp.yaml:225]: Thanks! I confess I didn't really study how these were used and had a feeling I was throwing away too much.

I'll add those back in, no problem. But I'm still not sure how they're used.. did I remove too much from configure_script_yaml.py ? Or are they used by a container specific tool?

File: fre/pp/ppyaml_test/pp.yaml (hunk 3 above)
Comment (singhd789) [reply fre/pp/ppyaml_test/pp.yaml:2]: True but we should definitely check. There is extra information in the schema.json now (since the updated pp.yaml would not have the other environment variables at the top)

Comment (singhd789) [reply fre/pp/ppyaml_test/pp.yaml:2]: And we should make sure everything is defined how it's supposed to be, either required or null, yadda yadda

File: fre/pp/ppyaml_test/pp.yaml (hunk 2 above)
Comment (singhd789) [reply fre/pp/ppyaml_test/pp.yaml:225]: No, you didn't get rid of too much regarding these! The configure script didn't actually use these yet.... They were used in the old `apply-edits` script I had but when we revamped that to be configure_script_yaml, I forgot to add these back in. I can probably work on adding the functionality for these in if we want. 

They were for a non-container use. For example, if one was on niagara:
- stage-history needed a `tmp` location set (this wasn't an issue on ppan because somewhere in environment things, it was set)
- then the cylc-run directory was taking up too much space (as the default installation was `home`; disk quota exceeded), so that install option would be populated with `--symlink-dirs........` (forgot the whole syntax at the moment) to point the cylc-run directory to be installed elsewhere (`/collab1` on niagara)

File: fre/pp/configure_script_yaml.py (hunk 1 above)
Comment (ceblanton) [reply fre/pp/configure_script_yaml.py:45]: I guess we can live with these for now.. bigger fish to worry about

File: fre/pp/ppyaml_test/pp.yaml (hunk 3 above)
Comment (ceblanton) [reply fre/pp/ppyaml_test/pp.yaml:2]: You were completely right- the schema needed some corresponding updates. I think I made them now. However, I'm still not sure about the user-defined variables (the new "FRE properties").

I removed all except &xyInterp, but it's not in schema and still seems to work. Must the user variable names be in the schema? I think it would be better if they didn't.

File: fre/pp/ppyaml_test/pp.yaml (hunk 2 above)
Comment (ceblanton) [reply fre/pp/ppyaml_test/pp.yaml]: I added them back. If they're at all useful, let's keep them for sure.

**Comment by ceblanton** [comment]
Thanks so much for your review @singhd789 . I updated the pp schema and added back the container-specific pp.yaml items. It's ready for another review from you. No rush.. next week is fine. thanks!

File: fre/pp/ppyaml_test/pp.yaml (hunk 3 above)
Comment (ceblanton) [reply fre/pp/ppyaml_test/pp.yaml:2]: Thanks you're right. I removed the rose-app and suite filepath configurations from the schema

File: fre/pp/ppyaml_test/pp.yaml (hunk 2 above)
Comment (ceblanton) [reply fre/pp/ppyaml_test/pp.yaml]: Definitely, let's keep them. I added them back in.

File: fre/pp/ppyaml_test/pp.yaml [hunk 6]
```diff
@@ -1,225 +1,87 @@
...
     do_refinediag: False
     do_atmos_plevel_masking: True
     do_preanalysis: False
     do_analysis: False
-    do_analysis_only: False
 
   directories:
     history_dir: "/archive/c2b/am5/2022.01/c96L65_am5f1a0r0_amip/gfdl.ncrc5-intel22-classic-prod-openmp/history"
-    pp_dir: "/archive/$USER/canopy/am5/c96L65_amip/pp"
+    pp_dir: "/archive/$USER/canopy/am5/c96L65_amip/gfdl.ncrc5-deploy-prod-openmp/pp"
```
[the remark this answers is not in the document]
ceblanton: It's a great question. I agree with you it's not quite right yet- too much duplication.

And completely, agreed- we must document how to create a pp.yaml from scratch, with those oddies.

fre/pp/ppyaml_test/pp.yaml
```diff
@@ -1,225 +1,87 @@
...
 tmpdir:
   tmpdirpath:
-
+￼
 install-exp-script:
-    - path: 
+    - path:
     - install-option:
-        install: 
+        install:
```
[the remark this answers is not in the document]
singhd789: I'm pretty sure you didn't change anything, which was why I was confused as to why it gave me an error message, haha

PR #95, opened 2024-05-21
Fix probably typo in including cylc-uiserver in meta.yaml

ceblanton:


ceblanton:
@bcc2761 thanks for adding the `cylc-uiserver` to meta.yaml.  It didn't seem to stick for some reason in the auto-deployed fms user environment. I saw that there is one colon where the others had two. Maybe it makes a difference?

bcc2761 (approved):

bcc2761:
oops...yeah. This should ideally fix it

Issue #99, opened 2024-06-03
Expand experiment, platform, target in pp.yaml definitions


ceblanton:
Currently, we specify directories simply, e.g.

```
  directories:
    history_dir: "/archive/c2b/am5/2022.01/c96L65_am5f1a0r0_amip/gfdl.ncrc5-intel22-classic-prod-openmp/history"
    pp_dir: "/archive/$USER/canopy/am5/c96L65_amip/gfdl.ncrc5-deploy-prod-openmp/pp"
    ptmp_dir: "/xtmp/$USER/ptmp"
    analysis_dir: "/nbhome/$USER/canopy/analysis/am5_c96L65_amip"
```

Bronx FRE directories allow specification in terms of the experiment, platform, target, stem, and others: see

https://sites.google.com/noaa.gov/oar-gfdl-msd-docs/fre-documentation/fre-documentation/xml-files?authuser=0#h.mmnsmti49ogn

e.g. the above directories could be specified more generically:

```
<directory>
    <archive>/archive/c2b/$(stem)/$(name)/$(platform)-$(target)/history</archive>
    <postProcess>/archive/$USER/canopy/$(stem)/$(name)/($platform)-$(target)/pp</postProcess>
    <ptmp>/xtmp/$USER/ptmp</ptmp>
    <analysis>/nbhome/$USER/canopy/analysis/$(name)</analysis>
</directory>
```

We want a similar "FRE-defined" variable expansion capability for the pp.yamls. e.g. how about?

```
  directories:
    history_dir: "/archive/c2b/am5/2022.01/$(expname)/$(platform)-$(target)/history"
    pp_dir: "/archive/$USER/canopy/am5/$(expname)/$(platform)-$(target)/pp"
    ptmp_dir: "/xtmp/$USER/ptmp"
    analysis_dir: "/nbhome/$USER/canopy/analysis/$(expname)"
```

So that users can then use the -e/-p/-t options to `fre pp configure-yaml` to have the same effect, e.g.

`fre pp configure-yaml --yaml=path/to/pp.yaml --experiment=c96L65_am5f1a0r0_amip --platform=gfdl.ncrc5-intel22-classic --target=prod-openmp`

singhd789:
Thanks for opening this up Chris. I agree. I think it might work but definitely need to see it in action

singhd789:
TO-DO: update schema.json

PR #101, opened 2024-06-12
 #99 Make pp yaml more generic

singhd789:
- update configure script to set experiment, platform, and target from click options


singhd789:
A lot of the changes are format fixes (pylint errors)

fre/pp/configure_script_yaml.py
```diff
@@ -1,125 +1,142 @@
...
+    for key,value in y.items():
+        if key == "rose-suite":
+            for suiteconfiginfo,dicts in value.items():
+                for configkey,configvalue in dicts.items():
+                    if configvalue is not None:
+                        k=configkey.upper()
+                        if k in ("HISTORY_DIR", "PP_DIR", "ANALYSIS"):
+                            # replace generic variables in pp yaml
+                            stem=e.split("_",1)[0]
+                            exp=e.split("_",1)[1]
```
ceblanton, line 86: This isn't safe. We can't safely assumes $(stem) is the first part of the experiment name, divided by an underscore.

Experiment, platform, and target are a special case for "FRE property" expansion as they are provided on the command line. "stem" is more of a regular "FRE property" in that it is just a regular variable name that we want to use in the pp.yamls

fre/pp/configure_script_yaml.py
```diff
@@ -1,125 +1,142 @@
...
+                            replacevars=configvalue.replace("$(stem)",stem).replace("$(expname)",exp).replace("$(platform)",p).replace("$(target)",t)
```
ceblanton, line 87: Yep, looks good. The first-draft expansion of Canopy "FRE properties".

Except we can't expand "stem" until we have a place to retrieve it from (properties.yaml?), and I suggest using "name" instead of "expname" as that is exactly as Bronx FRE Properties named them (so we have a bridge to Bronx-like behavior..)

fre/pp/ppyaml_test/pp.yaml
```diff
@@ -21,10 +21,10 @@ rose-suite:
     do_analysis: False
 
   directories:
-    history_dir: "/archive/c2b/am5/2022.01/c96L65_am5f1a0r0_amip/gfdl.ncrc5-intel22-classic-prod-openmp/history"
-    pp_dir: "/archive/$USER/canopy/am5/c96L65_amip/gfdl.ncrc5-deploy-prod-openmp/pp"
+    history_dir: "/archive/c2b/$(stem)/2022.01/$(expname)/$(platform)-$(target)/history"
+    pp_dir: "/archive/$USER/canopy/$(stem)/$(expname)/$(platform)-$(target)/pp" 
     ptmp_dir: "/xtmp/$USER/ptmp"
-    analysis: "/nbhome/$USER/canopy/analysis/am5_c96L65_amip"
+    analysis: "/nbhome/$USER/canopy/analysis/$(expname)"
```
ceblanton, line 27: Suggest "name" instead

fre/pp/ppyaml_test/pp.yaml
```diff
@@ -21,10 +21,10 @@ rose-suite:
     do_analysis: False
 
   directories:
-    history_dir: "/archive/c2b/am5/2022.01/c96L65_am5f1a0r0_amip/gfdl.ncrc5-intel22-classic-prod-openmp/history"
-    pp_dir: "/archive/$USER/canopy/am5/c96L65_amip/gfdl.ncrc5-deploy-prod-openmp/pp"
+    history_dir: "/archive/c2b/$(stem)/2022.01/$(expname)/$(platform)-$(target)/history"
+    pp_dir: "/archive/$USER/canopy/$(stem)/$(expname)/$(platform)-$(target)/pp"
```
ceblanton, line 25: In this case, "stem" wants to be "am5/2022.01". We can't do this, yet. So we could probably leave the hard-coded stem in the pp_dir:

pp_dir: "/archive/$USER/canopy/am5/2022.01/$(name)/$(platform)-$(target)/pp"

fre/pp/ppyaml_test/pp.yaml
```diff
@@ -21,10 +21,10 @@ rose-suite:
     do_analysis: False
 
   directories:
-    history_dir: "/archive/c2b/am5/2022.01/c96L65_am5f1a0r0_amip/gfdl.ncrc5-intel22-classic-prod-openmp/history"
-    pp_dir: "/archive/$USER/canopy/am5/c96L65_amip/gfdl.ncrc5-deploy-prod-openmp/pp"
+    history_dir: "/archive/c2b/$(stem)/2022.01/$(expname)/$(platform)-$(target)/history"
```
ceblanton, line 24: We really want "/archive/c2b/$(stem)/$(name)/$(platform)-$(target)/history", with stem being "am5/2022.01" but we can't expand "stem" yet.

ceblanton (changes requested):
Good as an interim FRE Properties-like implementation, but it's not safe to assume an underscore in the experiment name, and "stem" is really an arbitrary variable that we need to find a new properties.yaml file for perhaps.

The "expname" / "name" thing is more of an opinion. feel free to ignore that one..

fre/pp/configure_script_yaml.py
singhd789: I agree. I'm working on changing this after our discussion yesterday.

fre/pp/configure_script_yaml.py
```diff
@@ -12,6 +12,30 @@
...
+                # if the experiment listed in the click options matches 
+                # an experiment name listed in the main yaml, return 
+                # that experiment yaml
+                if i.get("expname") == experiment:
+                    # Load this specific pp yaml for rose suite and whatnot
+                    expyaml=i.get("ppyaml")
+
+    # return experiment yaml and defined FRE properties 
+    return(expyaml,stem)
+
```
ceblanton, line 38: This is great! FRE Property expansion happens here.

I think there has to be special case expansions for the command-line arguments: experiment, platform, and target: $(name), $(platform), and $(target).

For the rest, it would be ideal to have them be completely user-specified, and not limited to a particular set, e.g. "$(stem).

fre/pp/ppyaml_test/am5-yamls.yaml
```diff
@@ -0,0 +1,12 @@
+## FRE properties
+projectID: "am5"
+FRE_VERSION: &fre_version "canopy"
+AM5_VERSION: &am5_version "2022.01" #"am5f7b12r1"
+stem: "am5/2022.01" # stem=projectID/version
```
ceblanton, line 5: What do you think about organizing the properties under a header instead, like:

```
properties:
  projectID: "am5"
  FRE_VERSION: &fre_version "canopy"
  AM5_VERSION: &am5_version "2022.01"
  stem: "am5/2022.01"
```

In those 4 properties, projectID and stem have just one value, and FRE_VERSION and AM5_VERSION have another definitions, `&fre_version` and `&am5_versions`. Why the difference, and what is the `&` for? (Pardon my yaml ignorance..)

fre/pp/ppyaml_test/am5-yamls.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+# Experiments
+exp:
+  - expname: "c96L65_am5f1a0r0_amip"
+    platform: "gfdl.ncrc5-intel22-classic"
+    target: "prod-openmp"
+    ppyaml: "pp_c96L65.yaml"
```
ceblanton, line 12: I like this, mostly. So one does a

fre pp wrapper -y am5-yamls.yaml -p gfdl.ncrc5-intel22-classic -t prod-openmp -e c96L65_am5f1a0r0_amip

and first the am5-yamls.yaml is read, and then the pp_c96L65.yaml is also read. Yes?

I like this as is, but have some reservations about `target` and `platform`. For postprocessing, the target and platform are mostly used for directories, while the experiment is truly specific to the pp_c96L65.yaml settings.

Since we're passing the platform, target, and experiment to the `fre` tools, I think it's unnecessarily redundant to store them in the yaml also.

fre/pp/ppyaml_test/pp_c96L65.yaml
```diff
@@ -21,10 +21,10 @@ rose-suite:
     do_analysis: False
 
   directories:
-    history_dir: "/archive/c2b/am5/2022.01/$(name)/$(platform)-$(target)/history"
-    pp_dir: "/archive/$USER/canopy/am5/2022.01/$(name)/$(platform)-$(target)/pp" 
+    history_dir: "/archive/c2b/$(stem)/$(name)/$(platform)-$(target)/history"
+    pp_dir: "/archive/$USER/canopy/$(stem)/$(name)/$(platform)-$(target)/pp" 
     ptmp_dir: "/xtmp/$USER/ptmp"
-    analysis: "/nbhome/$USER/canopy/analysis/$(name)" 
+    analysis: "/nbhome/$USER/canopy/analysis/$(name)"
```
ceblanton, line 27: I have a strong feeling that `directories` should be set above the experiments, along with the FRE properties in the "main" yaml.

There's nothing in these definitions that are really experiment-specific except the experiment name itself.

ceblanton (changes requested):
Thank you! I think this is almost just what we need for AM5 runtime integration. Let's discuss..

fre/pp/ppyaml_test/am5-yamls.yaml
singhd789: Yes! The `am5-yamls.yaml` would be passed with the `-y` option in the command, as well as the `-e ...` option. So then, if the experiment passed as a click option matches the expname listed in the am5-yamls, it uses the information associated with that experiment name. From your example, 
`fre pp wrapper -y am5-yamls.yaml -p gfdl.ncrc5-intel22-classic -t prod-openmp -e c96L65_am5f1a0r0_amip`

1. `am5-yamls.yaml` is read
2. `-e c96L65_am5f1a0r0_amip` is given so it looks through the expname section of the yaml to match the experiment name given
3. uses the info associated with c96L65_am5f1a0r0_amip, being that specific experiment's ppyaml

I completely agree on the target and platform Chris! I thought about it more this morning, completely redundant and unnecessary.

fre/pp/ppyaml_test/pp_c96L65.yaml
singhd789: Aren't the directories experiment specific though with the name? I feel like it makes sense to keep them in the experiment's ppyaml because they would be paths for that experiment.

singhd789: Or at least putting them under the associated `expname` field

ceblanton: The best argument for keeping it in the main yaml is minimizing duplication. Almost certainly, all the directories will be the same, except for the experiment name, stem, platform, and target.

Another argument is similarity to how Bronx handles directories. In Bronx, the directories are set per-platform, not per-experiment.

fre/pp/configure_script_yaml.py
```diff
@@ -1,124 +1,256 @@
...
+                                  experiment=e,
+                                  platform=p,
+                                  target=t)
+
+    # Load the combined yaml
+    final_yaml = yaml_load(comb_yaml)
+
+    # Clean combined yaml to validate
+    del final_yaml["fre_properties"]
+    del final_yaml["shared"]
```
ceblanton, line 218: It's possible (but unlikely) that pp yamls could have no `fre_properties` or `shared` sections, do you agree?

If those two keys don't exist, then the `del` will cause an exception.

I think a `experiments` section is always required, though.

fre/pp/schema.json
```diff
@@ -3,86 +3,61 @@
...
-        "path": {"type": "string"},
-        "install-option": {
-          "type": "object",
-          "properties": {
-            "install": {"type": "string"}
-          }
-        }
-      }
-    }
+       "name": {"type": "string"},
```
ceblanton, line 6: We want this schema to be a submodule of https://github.com/NOAA-GFDL/gfdl_msd_schemas soon, but since this works, I think we should take this as is and refactor in another PR.

ceblanton (approved):
@singhd789 thank you so much for this.

Can this be merged in now? It's functional, though there will be needed updates as we use it. I suggest we merge and make more updates as needed.

fre/pp/schema.json
singhd789: I agree

fre/pp/configure_script_yaml.py
singhd789: Hmm I suppose there could be that case. I could add an check for that.

singhd789:
@ceblanton oh yep! I think it can be merged in, I forgot it was still in draft mode

PR #103, opened 2024-06-14
45.call.wo.decorators

cwhitlock-NOAA:
The pp wrapper is not ready for anything except testing (it's lacking the logic that we need for production work), but there are several smaller changes to the pp functions that I'd like to incorporate. These center around a wrapper/main function separation for the pp functions that allows you to call them without the decorators that click uses, which means you can invoke those functions from within python rather than on the command line.


cwhitlock-NOAA:
Apologies - I was not clean in which branch I was working on; there are some commits in checkoutScript that should not yet be in the repo. I'm removing the commits, re-merging and pushing.

bcc2761:
> Apologies - I was not clean in which branch I was working on; there are some commits in checkoutScript that should not yet be in the repo. I'm removing the commits, re-merging and pushing.

Ok, hopefully that resolves this error: https://github.com/NOAA-GFDL/fre-cli/actions/runs/9518276889/job/26238810991#step:4:1047

Issue #109, opened 2024-06-27
Use git tag for fre-cli versioning


ceblanton:
https://github.com/jbweston/miniver

To save us from manually updating the version in setup.py.

ilaflott:
To sum up discussion with @ceblanton today-

git tags will have the format `YYYY.#`, e.g. `2024.01`. When we decide to create a new git tag, e.g. `2024.02`, we wish for an approach to getting the version # *automatically updated* for publishing the conda package within github's CI/CD.

ilaflott:
this issue has been addressed

PR #111, opened 2024-06-27
draft: Use git tag for python/conda package version

ceblanton:
Followed instructions on

https://github.com/jbweston/miniver

may resolve https://github.com/NOAA-GFDL/fre-cli/issues/109

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+fre/_static_version.py export-subst
--- a/.github/workflows/build_conda.yml
+++ b/.github/workflows/build_conda.yml
@@ -11,6 +11,9 @@ jobs:
     steps:
     - name: Checkout Files
       uses: actions/checkout@v4
+      with:
+        fetch-depth: 1
+        fetch-tags: true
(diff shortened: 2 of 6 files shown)


ilaflott:
What do we want to accomplish here?

Iterate version number every merge to `main`?

ceblanton:
I think we want two deploying pipelines. One for versioned releases, that is triggered by creation of a new git tag. And the second for the testing release, that is triggered by pushes to `main`.

The part that I'd like from this PR is connecting the conda package version to the git tag version.

It doesn't work, for some reason. I'll change to draft and close it soon. I think replacing `setuptools` with `setuptools-scm` might be the most palatable solution to connect the conda package version to the git tag version.

ilaflott:
are these PRs supposed to be drafts? are they still being worked on? the title having "draft" does not make this PR a draft.

ilaflott:
closing for lack of activity

ceblanton:
Thanks Ian. Agreed. It didn't work well enough to really propose it.

Issue #107, opened 2024-06-21
Most scripts need clean-up


ilaflott:
the `pylint` step in the [CI/CD pipeline shows this problem](https://github.com/NOAA-GFDL/fre-cli/actions/runs/9570450402/job/26385224366) (search the logs for `pylint`). 

Too much non-PEP compliance increases the difficulty of reading, as well as maintaining the code.

Almost anyone could pick any script, and simply start trying to address the `pylint` complaints. extra whitespaces are a low hanging fruit; so are long lines. 

On this note, do we need a `CONTRIBUTING.md` to discuss what the barrier-to-entry should be, python-quality wise? This issue can serve as a starting point for that dialogue.

ilaflott:
this issue is in progress in various forms, on a tool-by-tool basis

PR #112, opened 2024-06-28
Use git tag for package versioning

ceblanton:
Using versioneer this time

https://github.com/python-versioneer/python-versioneer/blob/master/INSTALL.md

#107


ceblanton:
Perplexingly, this has the same problem. It's coming up with an unknown tag.

0+untagged.1.g4ab1d02

ilaflott:
same Q as on other pull:

What do we want to accomplish here?

Iterate version number every merge to main?

ilaflott:
are these PRs supposed to be drafts? are they still being worked on? the title having "draft" does not make this PR a draft.

ilaflott:
closing for lack of activity

ceblanton:
Ditto, thanks. I think we have a better plan forward for versioning, anyway.

https://github.com/NOAA-GFDL/fms_yaml_tools/blob/main/fms_yaml_tools/__init__.py#L1

PR #104, opened 2024-06-14
45.call.wo.decorators

cwhitlock-NOAA:
Mark II: removed the updates from another branch that I'd mixed into the current branch, fixed a few typos.

The pp wrapper is not ready for anything except testing (it's lacking the logic that we need for production work), but there are several smaller changes to the pp functions that I'd like to incorporate. These center around a wrapper/main function separation for the pp functions that allows you to call them without the decorators that click uses, which means you can invoke those functions from within python rather than on the command line.

PR #115, opened 2024-07-09
Add back the click decorator for yamlInfo

ceblanton:
A few small fixes for #104


cwhitlock-NOAA (approved):
Yes, that makes sense- we still need the click.command() when calling from the command line.

PR #135, opened 2024-07-29
Create fre/cmor/README.md

ilaflott:
## Describe your changes
adds back in old notes from original author's usage into `cmor` specific README.

no functionality changes. pipelines will be cancelled.


ceblanton:
I'm not sure these old notes are helpful... @Ciheim what do you think? We need some fresh cmor usage docs anyway

ceblanton:
On second thought, if those examples were rephrased in fre-cli syntax (fre cmor run blah blah) and showed how to get the input files (instead of assuming they are there), it would be very useful.

ilaflott:
**_> On second thought, if those examples were rephrased in fre-cli syntax (fre cmor run blah blah) and showed how to get the input files (instead of assuming they are there), it would be very useful._**

this is sort of what i was thinking. Also- i had actually meant to put these notes in `fre cmor`'s README in the first place, when i was doing some clean up items suggested by `pylint`.

ceblanton:
@ciheim, after you return let's clean up Sergey's old usage examples with the new syntax, and explain how to get the input config files.

ilaflott:
ready to go

ceblanton (approved):
the usage syntax needs to be updated but otherwise it's useful

Issue #124, opened 2024-07-16
Sphinx doc builds fail with v7.4.4 and current pip installed themes


ilaflott:
**Describe the bug**
documentation step of workflow broken as of https://pypi.org/project/Sphinx/7.4.4/#history
error message is:
```
Theme error:
An error happened in rendering the page FAQ.
Reason: UndefinedError("'style' is undefined")
```

source of the problem is the pip installed templates: https://github.com/NOAA-GFDL/fre-cli/blob/28908cb257c403062cbd7f6e14fa71eb6a0b28cb/.github/workflows/documentation.yml#L16

templates are set here: https://github.com/NOAA-GFDL/fre-cli/blob/28908cb257c403062cbd7f6e14fa71eb6a0b28cb/docs/conf.py#L27C1-L28C1
we can set the template `html_theme` to `sphinx_rtd_theme` instead, but this breaks for the same reason

**To Reproduce**
run a `documentation` workflow-  see https://github.com/NOAA-GFDL/fre-cli/actions/runs/9912991769/job/27518230011

**Expected behavior**
documentation being successfully built

**Additional context**
this is a tracking issue- these templates and `sphinx` are out of our hands. The current hot fix is to pin `sphinx` to version `7.3.7` and earlier. `7.3.7` is last working installed version of sphinx for the workflow

ilaflott:
This is also a tracking issue. Reopening and linking #148.

ilaflott:
closing as of #148. this issue will be re-opened if this problem comes back up again. I am noticing a pattern...

PR #148, opened 2024-08-05
upgrade the version of sphinx-rtd-theme to fix "style is undefined"

ceblanton:
## Describe your changes
Our sphinx CI building suddenly stopped working with this error:

Reason: UndefinedError("'style' is undefined")

Others have found that the cause is a mismatch between sphinx and sphinx-rtd-theme, and is fixed by updating the latter.

https://github.com/readthedocs/sphinx_rtd_theme/issues/1571

- [ ] I ran my code

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -14,9 +14,10 @@ jobs:
       - name: Install dependencies
         run: |
           #conda install sphinx
-          pip install sphinx renku-sphinx-theme sphinx-rtd-theme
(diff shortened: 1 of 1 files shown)


ceblanton:
#124

ilaflott:
I think we should be staying consistent between `pip` and `conda` in a way that seems intentional and non-arbitrary. I prefer the current version.

ilaflott:
no one-liner :-( but consistent

ilaflott (approved):

Issue #149, opened 2024-08-07
fre catalog build requires API update


ceblanton:
**Describe the bug**
Recent updates in CatalogBuilder,

https://github.com/NOAA-GFDL/CatalogBuilder/pull/29/files

require small updates to the fre-cli code.

**To Reproduce**
run `fre catalog build` on any pp directory

**Expected behavior**

**Additional context**

ceblanton:
@aradhakrishnanGFDL fyi

ilaflott:
I'm confused, the `click` API looks identical. This PR was closed, not merged?

ilaflott:
mm...
```
(fre_cli) an010: fre-cli $] fre catalog builder /archive/Ian.Laflotte/fre/FMS2023.04_om5_20240410/ESM4.2JpiC_om5b04r1/gfdl.ncrc5-intel23-prod-openmp/pp
Traceback (most recent call last):
  File "/home/Ian.Laflotte/conda/envs/fre_cli/bin/fre", line 8, in <module>
    sys.exit(fre())
  File "/home/Ian.Laflotte/conda/envs/fre_cli/lib/python3.9/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/home/Ian.Laflotte/conda/envs/fre_cli/lib/python3.9/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/home/Ian.Laflotte/conda/envs/fre_cli/lib/python3.9/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/Ian.Laflotte/conda/envs/fre_cli/lib/python3.9/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/Ian.Laflotte/conda/envs/fre_cli/lib/python3.9/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/Ian.Laflotte/conda/envs/fre_cli/lib/python3.9/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/home/Ian.Laflotte/conda/envs/fre_cli/lib/python3.9/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
[...]

PR #150, opened 2024-08-07
Update test_fre_cli.py

ilaflott:
`catalogbuilder`'s API changed, causing `fre catalog builder` to throw an error. the tests should catch this but do not currently.

updating with a test that will fail, which should show itself in the pipeline.

### DIFF
--- a/fre/tests/test_fre_cli.py
+++ b/fre/tests/test_fre_cli.py
@@ -58,12 +58,8 @@ def test_cli_fre_catalog():
     result = runner.invoke(fre.fre, args=["catalog"])
     assert result.exit_code == 0
 
-def test_cli_fre_catalog_builder():
-    result = runner.invoke(fre.fre, args=["catalog", "builder"])
-    assert result.exit_code == 0
-
-def test_cli_fre_catalog_builder():
-    result = runner.invoke(fre.fre, args=["catalog", "builder", "/archive/FIRST.LAST/fre/FMS2023.04_om5_20240410/ESM4.2JpiC_om5b04r1/gfdl.ncrc5-intel23-prod-openmp/pp"])
+def test_cli_fre_catalog_builder_help():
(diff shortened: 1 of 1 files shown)


ilaflott:
hooray! we have a test that will catch this now: 

https://github.com/NOAA-GFDL/fre-cli/actions/runs/10291762258/job/28484761124

#149

ilaflott:
normally i would say do-not merge because the tests are failing, but this is an exception 🤣. @Ciheim is working on the fix- i just wanted to make sure the test we define fails under the right conditions.

cwhitlock-NOAA (commented):
It looks good to me! This sounds like the tests caught exactly the kind of thing that they should catch.

ceblanton (approved):
hooray! Nice one.

ilaflott:
This actually was not complete. i should have put it in draft mode, that was my mistake.

This test was designed to catch a current issue (`CatalogBuilder` CI update), the name of the branch, "test_test_should_fail" indicated it's pipeline should have failed, but in fact, it succeeded. ergo, this code will not catch future errors, yet, and wasn't really ready for merge.

I'm not sure what kind of real review is being given to PRs, but i'd appreciate it if we all had a more careful eye

cwhitlock-NOAA:
About line 155 on this page sounds like it describes that behavior:

https://github.com/NOAA-GFDL/fre-cli/actions/runs/10291762258/job/28484761124

> =========================== short test summary info ============================
> FAILED fre/tests/test_fre_cli.py::test_cli_fre_catalog_builder - assert 1 == 0
>  +  where 1 = <Result AttributeError("module 'catalogbuilder.scripts.gen_intake_gfdl' has no attribute 'main'")>.exit_code
> ======== 1 failed, 64 passed, 2 skipped, 8 warnings in 85.71s (0:01:25) ========
> Error: Process completed with exit code 1.

Is there something else that I should have been looking for?

PR #151, opened 2024-08-09
These tests should NOT fail

ilaflott:
## Describe your changes
split up cli tests, add a new one based on the `catalogbuilder` interface, so we'll know if it changes again via a failed test. 

#149 

- [x] I ran my code
N/A ~~- [x] I wrote new instructions/documentation, if applicable~~


ilaflott:
this PR now contains autogenerated documentation from python doc-strings!
https://github.com/NOAA-GFDL/fre-cli/actions/runs/10372226899/job/28714459606?pr=151

ilaflott:
ready to go

fre/app/mask_atmos_plevel.py
```diff
@@ -49,61 +51,59 @@ def maskAtmosPlevel_subtool(infile, outfile, psfile):
...
     # transpose dims like the original array
     masked = masked.transpose(*ds[var].dims)
-
     print(f"Processed {var}")
 
     return masked
 
 
-def pressure_coordinate(ds, varname, verbose=False):
+def pressure_coordinate(ds, varname):#, verbose=False):
```
ceblanton, line 96: `verbose` wasn't used, I see

fre/catalog/frecatalog.py
```diff
@@ -1,15 +1,20 @@
+'''
+entry point for fre catalog subcommands
+'''
+
 import click
-import catalogbuilder
+#import catalogbuilder
```
ceblanton, line 6: Isn't this needed?

ilaflott: it is not! if it were, the new API tests in `fre/catalog/tests` would be failing

fre/catalog/frecatalog.py
```diff
@@ -25,17 +30,20 @@ def catalogCli():
 @click.pass_context
 def builder(context, input_path = None, output_path = None, config = None, filter_realm = None,
             filter_freq = None, filter_chunk = None, overwrite = False, append = False):
+    # pylint: disable=unused-argument
     """ - Generate .csv and .json files for catalog """
-    context.forward(gen_intake_gfdl.main)
+    context.forward(gen_intake_gfdl.create_catalog_cli)
```
ceblanton, line 35: this is the update!

fre/catalog/frecatalog.py
ceblanton: I can't disagree with you. But it's weird still isn't it? should we remove the line then?

fre/app/mask_atmos_plevel.py
ilaflott: true! commented out... since it's not a BAD idea

fre/catalog/frecatalog.py
ilaflott: and and and and 

https://github.com/NOAA-GFDL/fre-cli/blob/d237b690cdd4e593cbdea2202ea2ec7578baba47/fre/tests/test_fre_catalog_cli.py

ceblanton (approved):
Yeah! Lots of bonus good in this one, as well as the intended one line fix for the catalog builder api change.

PR #163, opened 2024-08-20
161.create combine yaml tool

singhd789:

- [x] I ran my code


fre/yamltools/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+### **Tests**
+
+To run `fre yamltools` test scripts, return to root directory of the fre-cli repo and call those tests with
+
+    python -m pytest fre/yamltools/tests/[test script.py]
+
+Or run all tests with
+
+    python -m pytests fre/yamltools/tests
```
ilaflott, line 13: `pytests` --> `pytest`

fre/yamltools/combine_yamls.py
```diff
@@ -0,0 +1,265 @@
...
+
+    if experiment not in exp_list:
+        raise NameError(f"{experiment} is not in the list of experiments")
+
+    # set platform yaml filepath
+    if comb_model["shared"]["compile"]["platformYaml"] is not None:
+        py=comb_model["shared"]["compile"]["platformYaml"]
+        py_path=Path(os.path.join(mainyaml_dir,py))
+    else:
+        py=False
```
ilaflott, line 82: seems unused?

fre/yamltools/combine_yamls.py
```diff
@@ -0,0 +1,265 @@
...
+    """
+    Combine compile yaml with the defined combined.yaml
+    Arguments:
+        comb_m       :  combined model yaml file
+        compileyaml  :  compile yaml file 
+    """
+    combined = comb_m
+
+    # copy compile yaml info into combined yaml
+    if compileyaml is not False:
```
ilaflott, line 123: worth noting in this function if `compileyaml` is a string representing a path to a `yaml`, or if it is a loaded `yaml` object itself... or perhaps, it's a boolean?

Since later lines imply `compileyaml` is a string, `if compileyaml is not None:` will be preferable.

fre/yamltools/combine_yamls.py
```diff
@@ -0,0 +1,265 @@
...
+def combine_platforms(comb_mc,platformsyaml):
+    """
+    Combine platforms yaml with the defined combined.yaml
+    Arguments:
+        comb_mc        : combined model and compile yaml file
+        platformsyaml  :  platforms yaml file
+    """
+    combined = comb_mc
+    # combine platform yaml
+    if platformsyaml is not False:
```
ilaflott, line 139: see prev comment- is this pattern here for a GOOD reason? or is this a propagated issue?

if it's for a GOOD reason, that reason should be noted down somewhere nice and visible. otherwise we should change this to `XXXyaml is not None:` wherever that applies

fre/yamltools/tests/test_combine_yamls.py
```diff
@@ -0,0 +1,59 @@
...
+    assert Path(f"{CWD}/{yamls_dir}/yaml_include/analysis.yaml").exists()
+
+def test_merged_yamls():
+    """
+    Check for the creation of the combined-[experiment] yaml 
+    Check that the model yaml was merged into the combined yaml
+    """
+    os.chdir(in_dir)
+
+    modelyaml = Path(f"{CWD}/{yamls_dir}/am5.yaml")
```
ilaflott, line 55: looking at the top- can we remove `os.chdir` here in favor of making the later line, equivalently, `modelyaml = Path(f"{in_dir}/{yamls_dir}/am5.yaml")`?

If we can't b.c. this makes the test fail, i'd call that a fragility that is worth exploring. Either as part of this PR, or opening a sep issue as needed

fre/yamltools/combine_yamls.py
singhd789: meant to change `py=False` to `py_path=False`

fre/yamltools/combine_yamls.py
singhd789: Initially, I had these as `is not None` but I noticed issues with the analysis script merge section. The analysis scripts are not fully developed yet but we think they will act the same as  the experiment pp yamls (listed under each experiment if created). Soo, when the script looks for `analysisyaml=i.get("analysis")` that isn't there yet, it would say 
```
UnboundLocalError: local variable 'ay_path' referenced before assignment
```
Paired with a pylint message that I can't even remeber anymore, I think at one point I thought I had to set the if statement to something like `if analysisyaml not defined, it's false`. Big oof on my part, I can just set it to `None` instead of `False` and keep what I originally had with `is not None`. 

I'll fix this, thanks!

fre/yamltools/tests/test_combine_yamls.py
ilaflott: wow- it did break the test didn't it? not fun

I'm now worried about this too: https://github.com/NOAA-GFDL/fre-cli/blob/main/fre/pp/tests/test_configure_script_yaml.py#L30-L53

ilaflott: If this works fine with the `os.chdir` there, we can merge it in. I can open a separate issue for this odd dependence on `os.chdir` in different `fre yaml` routines

singhd789: lol, yeah although I thought maybe it was related to some path in the action. Like I thought I've seen this before but I basically made this test similar to the configure_script_yaml one. So now I'm trying to troubleshoot what happened and how things broke.

singhd789: If I can't figure it out today, I can add that `os.chdir` back in

singhd789:
@ilaflott the reason the build was failing was because I literally forgot a `( )`. I think it should be fine now if you want to re-review.

ilaflott (approved):
Well done @singhd789 ! Happy to approve+merge this one

PR #181, opened 2024-09-12
Fre pp tests tweak

ilaflott:
this is a quick hit while i'm waiting for a conda environment to build!

this PR takes the solid first pass effort at `fre pp` cli-calls and tweaks them into the current pytesting approach.
changes:
- split the argument list to `CliRunner` with space as the delimeter
- change argument order to `CliRunner`
- adds a base CLI call with no explicit flag(s)


ceblanton (approved):
Nice one. thank you!

PR #192, opened 2024-09-27
Update "fre catalog builder" test after upstream CatalogBuilder change

ceblanton:
Previously, the test verified the exit code and message for

(fre-cli) c2b:~/git/fre-cli%>fre catalog builder
No paths given, using yaml configuration

(fre-cli) c2b:~/git/fre-cli%>echo $status
1

Recently, the message was updated to be:

(fre-cli) c2b:~/git/fre-cli%>fre catalog builder 
Missing: input_path or output_path. Pass it in the config yaml or as command-line option

(fre-cli) c2b:~/git/fre-cli%>echo $status 
1

So let's continue to check for the exit code and not require an exact error message.

- [x] I ran my code

### DIFF
--- a/fre/tests/test_fre_catalog_cli.py
+++ b/fre/tests/test_fre_catalog_cli.py
@@ -24,12 +24,7 @@ def test_cli_fre_catalog_opt_dne():
 def test_cli_fre_catalog_builder():
     ''' fre catalog builder '''
     result = runner.invoke(fre.fre, args=["catalog", "builder"])
-    assert all( [
-                  result.exit_code == 1,
-                  'No paths given, using yaml configuration'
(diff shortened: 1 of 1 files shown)


fre/tests/test_fre_catalog_cli.py
ilaflott, line 1: The text gets checked because other problems will throw an exit code of 1, but a different error message.

ilaflott (changes requested):

ilaflott (approved):

fre/tests/test_fre_catalog_cli.py
ceblanton: Thanks! OK, I see. Changed back.

PR #198, opened 2024-10-01
Pylint edits

ilaflott:
removed unnecessary import statements from `__init__.py` files, update `pylint` call to give summative information in addition to every last little complaint.

addressed a few different pylint complaints, superfluous parents, some long lines, some quick `whitespace-cleanup` calls. fix indentation for one script in particular.

freezing this here before i get too carried away again...


fre/make/runFremake.py
```diff
@@ -56,136 +61,146 @@ def fremake_run(yamlfile,platform,target,parallel,jobs,no_parallel_checkout,verb
...
-
-    ## TODO: Options for running on login cluster?
-                   freCheckout.run()
+        if modelYaml.platforms.hasPlatform(platformName):
+            pass
+        else:
+            raise ValueError(f'{platformName} does not exist in '
+                             f'{modelYaml.combined.get("compile").get("platformYaml")}')
+
+        ( compiler, modules, modulesInit,
```
singhd789, line 77: Thank you for fixing this some more!

fre/pp/configure_script_yaml.py
```diff
@@ -87,9 +87,9 @@ def quote_rose_values(value):
     boolean, in which case do not quote them.
     """
     if isinstance(value, bool):
-        return(f"{value}")
+        return f"{value}"
```
singhd789, line 90: Since it's just returning the value  now, I think we can remove the `f"..."` and just have `value`

fre/yamltools/data_table/combine_data_table_yamls.py
```diff
@@ -118,4 +118,4 @@ def combine_data_table_yaml(in_files, out_file, force):
 
 
 if __name__ == "__main__":
```
singhd789, line 120: Just FYI: I don't think the data_table and diag_table yamltools are really integrated in the fre-cli yet. 

They live here: https://github.com/NOAA-GFDL/fms_yaml_tools/tree/main/fms_yaml_tools

We added them in the fre-cli with the intent to add them in but I don't think anything else was done with them yet.

singhd789:
The pylint fixes look good!

fre/app/generate_time_averages/generate_time_averages.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
```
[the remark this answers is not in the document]
ilaflott: thinking it through- we want users to call this function at the command line, i would hope they use `fre app gen-time-averages`! the current click-based CLI usage is tested for.

we also facilitate using `generate_time_averages` as an importable module- this is tested for!

we have no intention of facilitating direct-execution not-via-click at this time, as i understand it. and we do not test for it's usage, but this is what the she-bang at the top facilitates. so, in order to avoid confusion in use cases, i think we remove it.

fre/yamltools/__init__.py
```diff
@@ -1,9 +0,0 @@
-''' for fre.yamltools imports '''
```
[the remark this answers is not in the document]
ilaflott: see tests, they all work!

the one interesting subcase that needed a non-empty `__init__.py` was `fre/cmor/__init__.py`. In this case, it lets one import `fre.cmor.cmor_run_subtool` instead of `fre.cmor.cmor_mixer.cmor_run_subtool`. i.e. convenience/brevity/namespace-stuff

fre/make/runFremake.py
ilaflott: f-strings are your friend and so are built-in exceptions!

fre/yamltools/data_table/combine_data_table_yamls.py
ilaflott: fair enough! we need to make `fms_yaml_tools` a proper dependency and call ala `catalogbuilder`

ilaflott:
> The pylint fixes look good!

thank you for the feedback! github doesn't consider this officially reviewed with approval/request-for-changes/null approval until you click the 'review changes' button, top-right of the 'files changed' tab on this PR

fre/pp/configure_script_yaml.py
singhd789: That's fine too, it works and it's not a problem having f-strings at all

singhd789 (approved):

PR #199, opened 2024-10-01
Add regrid xy

ilaflott:
inspired by https://github.com/NOAA-GFDL/fre-workflows/pull/32

- [x] I ran my code


fre/app/freapp.py
```diff
@@ -7,70 +7,118 @@
...
+              type = str,
+              help = "`outputDir` / `output_dir` (env var) specifies target location for output" + \
+                     " regridded files",
+              required = True)
+@click.option("-b", "--begin",
+              type = str,
+              help = "`begin` / `begin` (env var) ISO8601 datetime format specification for" + \
+                     " starting date of data, part of input target file name",
+              required = True)
+@click.option("-t", "--tmp_dir",
```
singhd789, line 32: It's pretty easy to use options like this (one letter) but I also feel like some of these can have more explicit or different option flags (the first `-[option]` at least). Maybe this could be `-tmp` because `-t` is used in other places too, referencing things like `time` or `target`. It could get confusing with `-t` options that refer to a bunch of things. Thoughts?

fre/app/freapp.py
```diff
@@ -7,70 +7,118 @@
...
+              type = str,
+              help = "`TMPDIR` / `tmp_dir` (env var) temp directory for location of file " + \
+                     "read/writes",
+              required = True)
+@click.option("-r", "--remap_dir",
```
singhd789, line 37: Similar comment to above, `-r` usually refers to recursive, consider even `-rd` for remap_dir? (may not be the best still)

fre/app/freapp.py
```diff
@@ -7,70 +7,118 @@
...
+              type = str,
+              help = "`source` / `source` (env var) source name for input target file name " + \
+                     "within input directory to target for regridding. the value for `source` " + \
+                     "must be present in at least one component's configuration fields",
+              required = True)
+@click.option("-g", "--grid_spec",
+              type = str,
+              help = "`gridSpec` / `grid_spec` (env var) file containing mosaic for regridding",
+              required = True)
+@click.option("-I", "--def_xy_interp",
```
singhd789, line 52: `-I` is sometimes used for other `includes` needed in a model. What about `-xy`? (still may not be best - just trying to come up with what other flags could be)

fre/app/freapp.py
```diff
@@ -7,70 +7,118 @@
...
+              help = "Input NetCDF file containing pressure-level output to be masked",
+              required = True)
 @click.option("-o", "--outfile",
-              type=str,
-              help="Output file",
-              required=True)
+              type = str,
+              help = "Output file",
+              required = True)
 @click.option("-p", "--psfile",
```
singhd789, line 73: `-p` is also often used for `platform`. What about `ps` or `-psf`?

fre/app/freapp.py
```diff
@@ -7,70 +7,118 @@
...
 @click.pass_context
 def mask_atmos_plevel(context, infile, outfile, psfile):
-    # pylint: disable=unused-argument
+    # pylint: disable = unused-argument
     """Mask out pressure level diagnostic output below land surface"""
     context.forward(mask_atmos_plevel_subtool)
 
 
 @app_cli.command()
 @click.option("-i", "--inf",
```
singhd789, line 84: Maybe, to keep consistent, also make `--inf` --> `--infile`

fre/app/freapp.py
```diff
@@ -7,70 +7,118 @@
...
-              type=str,
-              required=True,
-              help="Input file name")
+              type = str,
+              required = True,
+              help = "Input file name")
 @click.option("-o", "--outf",
```
singhd789, line 88: `--outf` to `--outfile`

fre/app/freapp.py
```diff
@@ -7,70 +7,118 @@
...
-              type=str,
-              required=True,
-              help="Output file name")
+              type = str,
+              required = True,
+              help = "Output file name")
 @click.option("-p", "--pkg",
```
singhd789, line 92: Similar comment to before, `-p` could be `platform`, `-p` could be `psfile`, etc. 

To make more clear 
`-p` --> `-pkg`
`--pkg` --> `--tavg-pkg`

fre/app/freapp.py
```diff
@@ -7,70 +7,118 @@
...
-              default="all",
-              help="Type of time average to generate. \n \
-                    currently, fre-nctools and fre-python-tools pkg options\n \
-                    do not support seasonal and monthly averaging.\n")
+              type = click.Choice(["month","seas","all"]),
+              default = "all",
+              help = "Type of time average to generate. \n \
+                     currently, fre-nctools and fre-python-tools pkg options\n \
+                     do not support seasonal and monthly averaging.\n")
 @click.option("-s", "--stddev_type",
```
singhd789, line 110: `-s` could also be `source` as defined from above. Maybe `-s` --> `-sd`?

fre/yamltools/combine_yamls.py
```diff
@@ -369,12 +366,12 @@ def _consolidate_yamls(yamlfile,experiment,platform,target,use):
         raise ValueError("'use' value is not valid; must be 'compile' or 'pp'")
 
 @click.command()
-def consolidate_yamls(yamlfile,experiment,platform,target,use):
+def _consolidate_yamls(yamlfile,experiment,platform,target,use):
```
ceblanton, line 369: Okay, so the plain name is able to be called from other methods (pythonically), and the underscore methods are used for the click interface?

ceblanton (approved):
Woohoo! Does it really work as-is even with the Rose config usage?

fre/make/createCompile.py
```diff
@@ -6,7 +6,7 @@
 from pathlib import Path
 from multiprocessing.dummy import Pool
 import click
-from .gfdlfremake import varsfre, platformfre, yamlfre, targetfre, buildBaremetal
+from .gfdlfremake import varsfre, yamlfre, targetfre, buildBaremetal
```
singhd789, line 9: I'm not sure about this. One of the many reasons we need to make some tests for fre make, to see if this still works. I think there's a bit in this script that references `.hasPlatforms` which I believe is a function in the `platforms` class in `platformfre.py`

fre/make/createDocker.py
```diff
@@ -4,7 +4,8 @@
 import sys
 from pathlib import Path
 import click
-from .gfdlfremake import varsfre, targetfre, makefilefre, platformfre, yamlfre, buildDocker
+#from .gfdlfremake import varsfre, targetfre, makefilefre, platformfre, yamlfre, buildDocker
+from .gfdlfremake import varsfre, targetfre, yamlfre, buildDocker
```
singhd789, line 8: Same comment as in `createCompile.py`

fre/yamltools/combine_yamls.py
singhd789: @cwhitlock-NOAA I thought the underscore method was something for the pp wrapper?

fre/make/createCompile.py
singhd789: Ok yeah, you can disregard this comment. This line in fremake that uses the `platforms` class and `hasPlatform` function 
```
if modelYaml.platforms.hasPlatform(platformName):
```
is an object (modelYaml) created using `yamlfre` (which uses platformfre). We're all good here!

fre/pp/configure_script_yaml.py
```diff
@@ -10,18 +10,17 @@
...
 import yaml
 import metomi.rose.config
 
 # Relative import
-f = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-sys.path.append(f)
-import yamltools.combine_yamls as cy
+#f = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+#sys.path.append(f)
+import fre.yamltools.combine_yamls as cy
```
singhd789, line 23: Thank you! I thought I fixed this already but apparently I forgot this one

fre/yamltools/combine_yamls.py
ilaflott: > Okay, so the plain name is able to be called from other methods (pythonically), and the underscore methods are used for the click interface?

CORRECT!



> @cwhitlock-NOAA I thought the underscore method was something for the pp wrapper?

and also, let me clear, @cwhitlock-NOAA introduced me to the idea. I think it just got used backwards in a spot or two. but conventionally to `pylint`, `_some_function_name` is the "protected" one... in our case, it's "protected" for `click`.

fre/app/freapp.py
ilaflott: if the syntax is valid `-xy` is definitely going to be better here, and i should apply it around real quick

fre/app/freapp.py
ilaflott: good catches!

fre/app/freapp.py
ilaflott: change implemented in next commit

fre/app/freapp.py
ilaflott: change implemented next commit

fre/app/freapp.py
ilaflott: change implemented next commit

fre/app/freapp.py
ceblanton: i agree with Dana. Bronx has long had an annoying ambiguity between -t being --target and --time, and it's worth keeping the the options consistent whereever possible.

How about we outlaw the short options altogether except in especially well-known cases?

fre/app/freapp.py
ilaflott: gonna leave this- i haven't touched that script and some slightly closer reading yields some inconsistency that makes me not want to further confuse it. see after `--psfile` ... "... file containing surfase pressure (ps)"... is that the case? or is it like a post-script image file?

in any case... to look into... marked TODO next commit

fre/app/freapp.py
ilaflott: that's `gen-time-averages` and would entail going through to change the args. if you don't mind, i'll leave that for now!

PR #202, opened 2024-10-03
remove unused fre/yamltools/[diag_table, data_table, field_table] from fre/yamltools

ilaflott:
If we remove this as-of-yet un-used, un-tested code, our coverage % post merge of this and the `add_regrid_xy` branch would be about 10% higher

any reason that we can't remove this code? who is using the versions of the scripts in this repo instead of those in [fms_yaml_tools](https://github.com/NOAA-GFDL/fms_yaml_tools)?


ceblanton (approved):
Yeah!

PR #205, opened 2024-10-11
cmor tool improvements

ilaflott:
- [x] I ran my code
- [x] I tried to make my code readable
- [x] I tried to comment my code
- [x] I wrote a new test, if applicable
- [x] I wrote new instructions/documentation, if applicable
- [x] I ran pytest and inspected it's output
- [x] I ran pylint and attempted to implement some of it's feedback

This PR...
- overhauls the `README.md` with `fre-cli` oriented usage and removes outdated doc
- refactors `cmor_mixer.py` heavily to be more readable, have more modularity to it to clearly illustrate who-does-what and why, with an emphasis on clearly exposing data movement inherent in the routine
- functions now have doc strings with argument descriptions in most places
- better run-time feedback of functioning/IO/logic etc during runtime, hopefully not too much
- better comments where possible
- new test case where `key:value` pair in `var_list` argument differ
- more test case cleanup/setup, more specific test case success conditions
- new CLI/`click.testing.CliRunner` test for the new functionality of the `fre.cmor` module
- change of the way the click module interfaces with the cmor run subtool module

PR #218, opened 2024-10-24
217 teaching `fre cmor` to bite

ilaflott:
this branches off of https://github.com/NOAA-GFDL/fre-cli/pull/205

gfdl-only specialty tests progress, the calls of which are seen in [run_test_file_cases.py](https://github.com/NOAA-GFDL/fre-cli/blob/217-the-cmorbite-one-variable8-grids/run_test_file_cases.py)
- [x] 1. land gr1 / Lmon / gr1
- [x] 2. atmos gr1 / Amon / cl (full vertical levels)
- [x] 3. atmos gr1 / Amon / mc (half vertical levels)
- [ ] 4. atmos gr1 / Amon / ta
    - [ ] allow no-lon coordinate axis case
- [x] 5. ocean gr / Omon / sos
- [ ] 6. ocean gn / Omon / sos
    - [ ] `nc_copy` bug
- [ ] 7. ocean gr / Omon / so (three dimensional ocean)
    - [ ] `nc_copy` bug
- [ ] 8. atmos gn / Amon / ch4global
    - [ ] allow no-lon coordinate axis case
    - [ ] allow no-lat coordinate axis case
- [ ] 9. LUmip gr1 Emon / gppLut
    - [ ] come up with an edge-case work around for non-typical coordinate in place of a more typical vertical coordinate

code QA:
- [x] I ran my code

This PR adds...
- fix for `tmp` directory + filename handling
- fix: `cmor_lev` no longer referenced before assignment
- doc string for `rewrite_netcdf_vars`
- a currently one-off gfdl-only testing script, `run_test_file_cases.py` outside of the pytest framework to avoid making the repo explode in size
- updates the directory `README` accordingly
- updates the click interface help menu read outs


ilaflott:
Pausing to ponder for a second- the following two commits are really about a different problem.
```
commit 31996b08b3c2ab3c5510a86606c01ef305383a55 (HEAD -> 217-the-cmorbite-one-variable8-grids, 
    fix undefined variable name errors

commit 4b313384a8c0cda3bb4eab2d398d99d3ada8a93c
    alrightly, lets move some of this hardcoding to the top of the file, see how it goes...

```
I could branch that development off here and revert the commits to make this less about exposing configuration and more about accommodating a wider array of input cases.

ilaflott:
spinning off [branch `fre-cmor-config-exposed`](https://github.com/NOAA-GFDL/fre-cli/tree/fre-cmor-config-exposed) at commit 
`31996b08b3c2ab3c5510a86606c01ef305383a55` from this PR. Then, this PR's branch will have the aforementioned commit, `4b313384a8c0cda3bb4eab2d398d99d3ada8a93c`, and possibly parts of `62852833ecc7f42c3fbae1433fbaf341bba469cb` reverted, as those edits are really attempting to do something worthwhile (making hard-coded dependencies more obvious) but not related to this PR's primary goal. 

for the exposed-config branch, see draft PR #227

ilaflott:
as of 434dc3e9aeda5d1bfa334df538ecd9f43a7c6839, 4 out of the 9 cases in `run_test_file_cases` work outright. of the five failures, two are due to the lack of a lon/lat/both coordinates, two are actually associated with the `copy_nc` routine, and the last is associated with 4-dimensional land-use data where one has the typical horizontal coordinates, time, and NOT a z-coordinate

ilaflott:
we should be using a variable's "bounds" attributes to pull in assoc coordinate data:
https://github.com/NOAA-GFDL/fre-cli/pull/205/files#r1821226599 

might as well look at `lat` and `lon` or other coordinates similary.

ilaflott:
[from PR #205 feedback](https://github.com/NOAA-GFDL/fre-cli/pull/205/files#r1821217096) "do we really need to copy the input file before CMORizing?" - @ceblanton 

i don't particularly think we need to, maybe batch related? --> perhaps batch configuration option... or check

ilaflott:
[from PR #205 feedback](https://github.com/NOAA-GFDL/fre-cli/pull/205/files#r1821215025) - regarding `copy_nc`, "If the input file is NC3 will the output file be NC3 or NC4"

need to consider file-format preservation/specification

ilaflott:
[from PR #205 feedback](https://github.com/NOAA-GFDL/fre-cli/pull/205/files#r1821223129) - regarding `create_tmp_dir`, the logic of this function should be clarified or revisited

ilaflott:
[from PR #205 feedback](https://github.com/NOAA-GFDL/fre-cli/pull/205/files#r1821226599) 

generally: we should be using a variable's "bounds" attributes to pull in assoc coordinate data. This came up discussing time axes, but might as well look at `lat` and `lon` or other coordinates similary.

ilaflott:
next PR #237 will address above todo cases 4, 6, 7, 8, and 9.

fre/cmor/README.md
```diff
@@ -29,18 +29,39 @@ this subtool's help, and command-specific `run` help:
...
+                           CMORization. this metadata is effectively appended 
+                           to the final output file's header  [required]      
+  -o, --outdir TEXT        directory root that will contain the full output   
+                           and output directory structure generated by the    
+                           cmor module upon request.  [required]              
+  -v, --opt_var_name TEXT  optional, specify a variable name to specifically  
+                           process only filenames matching that variable name.
+                           I.e., this string help target local_vars, not      
+                           target_vars.                                       
+  --help                   Show this message and exit.
```
ceblanton, line 64: I can't ever really disagree with documentation, and don't disagree with this, but we'll have to keep this up-to-date with the actual --help output content which is defined elsewhere. Eventually, documentation like this will be auto-created through sphinx somehow.

fre/cmor/cmor_mixer.py
```diff
@@ -56,22 +57,24 @@ def copy_nc(in_nc, out_nc):
...
-def get_var_filenames(indir, var_filenames = None):
+def get_var_filenames(indir, var_filenames = None, local_var = None):
     '''
-    appends files ending in .nc located within indir to list var_filenames accepts two arguments
+    appends files ending in .nc located within indir to list var_filenames accepts three arguments
         indir: string, representing a path to a directory containing files ending in .nc extension
         var_filenames: list of strings, empty or non-empty, to append discovered filenames to. the
                        object pointed to by the reference var_filenames is manipulated, and so need
                        not be returned.
+        local_var: string, optional, if not None, will be used for ruling out filename targets
```
ceblanton, line 67: What is the purpose of this `local_var` setting? Is it related to the optional name renaming (e.g. t_ref to tas)?

fre/cmor/cmor_mixer.py
```diff
@@ -141,20 +144,20 @@ def create_tmp_dir(outdir):
...
               outdir.find("/net" ) != -1 ] ):
-        print(f'(cmorize_target_var_files) using /local /work /net ( tmp_dir = {outdir}/ )')
-        tmp_dir = "{outdir}/"
+        print(f'(create_tmp_dir) using /local /work /net ( tmp_dir = {outdir}/ )')
+        tmp_dir = str( Path("{outdir}/").resolve() )
     else:
-        print('(cmorize_target_var_files) NOT using /local /work /net (tmp_dir = outdir/tmp/ )')
-        tmp_dir = f"{outdir}/tmp/"
+        print(f'(create_tmp_dir) NOT using /local /work /net (tmp_dir = {outdir}/tmp/ )')
+        tmp_dir = str( Path(f"{outdir}/tmp/").resolve() )
```
ceblanton, line 156: we agreed to revisit this tmpdir and if it's needed at all. It seems odd to combine the string and path type conversion in one line but I've had to do it too.

fre/cmor/cmor_mixer.py
```diff
@@ -166,12 +169,23 @@ def rewrite_netcdf_file_var ( proj_table_vars = None,
...
+    accepts six arguments, all required:
+        proj_table_vars: json dictionary object, variable table read from json_table_config.
+        local_var: string, variable name used for finding files locally containing target_var,
+                   this argument is often equal to target_var.
+        netcdf_file: string, representing path to intput netcdf file.
+        target_var: string, representing the variable name attached to the data object in the netcdf file.
+        json_exp_config: string, representing path to json configuration file holding metadata for appending to output
+                         this argument is most used for making sure the right grid label is getting attached to the right output
+        json_table_config: string, representing path to json configuration file holding variable names for a given table.
+                           proj_table_vars is read from this file, but both are passed anyways.
```
ceblanton, line 183: OK, so local_var is more clearly related to the variable renaming feature (t_ref => tas), I see.

fre/cmor/cmor_mixer.py
```diff
@@ -311,10 +322,11 @@ def rewrite_netcdf_file_var ( proj_table_vars = None,
...
                                        zfactor_values = ds["b_bnds"][:],
                                        units          = ds["b_bnds"].units )
-                cmor_lev = cmor.axis( "alternate_hybrid_sigma_half",
-                                      coord_vals = lev[:],
-                                      units = lev.units )
             else:
+                cmor_lev = cmor.axis( "alternate_hybrid_sigma",
+                                      coord_vals  = lev[:],
+                                      units       = lev.units,
+                                      cell_bounds = ds[vert_dim+"_bnds"] )
```
ceblanton, line 329: Is this line creating the cell bounds from scratch or do they have to be in the input file?

If it's creating it from thin air, then this should be fine. But if it has to be in the input file, then perhaps it is named something different, but is probably `vert_dim+'_bnds'`. The CF way to find it would be similar to the dimension axis:

	double lev(lev) ;
		lev:units = "1.0" ;
		lev:long_name = "hybrid sigma pressure coordinate" ;
		lev:axis = "Z" ;
		lev:positive = "down" ;
		lev:formula = "p(n,k,j,i) = ap(k) + b(k)*ps(n,j,i)" ;
		lev:formula_terms = "ap: ap b: b ps: ps" ;
		lev:bounds = "lev_bnds" ;
		lev:standard_name = "atmosphere_hybrid_sigma_pressure_coordinate" ;

ceblanton (approved):
Looks great, Ian. I didn't run it, but with your test cases I believe you, and we'll have plenty more opportunities in the future to run it.

I was slightly confused about the optional "local var" setting, mostly wondering why it's not required and normally set to the "target" variable name. But later on the "local var" IS required, so I think I just confused myself.

And then the question about whether the dimension bounds should be retrieved in a CF-friendly way, but regardless it's a fair assumption.

Aside from those two probable non-issues, all of the changes are clear improvements for functionality and clarity. Onward!

fre/cmor/README.md
ilaflott: I've been pondering this too. hear me out for a second- I know that the way we generate coverage badges in our pipeline is a bit hacky- utilizing the deployed github pages to host the generated image at a static URL for linking to the README. 

But... there DO exist workflows that edit the coverage badge in the repo directly and makes the commit itself- i tried to make those work but couldn't figure it out on a reasonable timeline.

Why do i bring this up? Well, in theory if a workflow could commit to the repo for a generated test-coverage badge, it should also be able to commit a quick README edit based on the latest `fre <tool> --help` or `fre <tool> <subtool> --help`

fre/cmor/cmor_mixer.py
ilaflott: yes! that's the idea as i've come to understand it, and initially needed to even be convinced of this by you! I tried to clarify this functionality with `fre/tests
/test_fre_cmor_cli.py::test_cli_fre_cmor_run_case2` and `fre/cmor/tests
/test_cmor_run_subtool.py::test_fre_cmor_run_subtool_case2`

fre/cmor/cmor_mixer.py
ilaflott: `ds` is grabbed as a `netCDF4.Dataset` from the targeted files in the input directory, so `ds[vert_dim+'_bnds']` demands a value pointed to by key `vert_dim + '_bnds'` exists in `ds`.

fre/cmor/cmor_mixer.py
ilaflott: see #240 ! but perhaps we should discuss the best way to use/handle `pathlib.Path` in our codebase... lest we get too deep in the hole!

fre/cmor/cmor_mixer.py
ilaflott: #241

ilaflott:
case 4 works as of #243 / [5fb52ce](https://github.com/NOAA-GFDL/fre-cli/pull/243/commits/5fb52ceb804edf8d2355024b7f06282e403808e3)

PR #231, opened 2024-10-29
 230.Add decorator functions to pass click options correctly

singhd789:
- this was needed in order to import fre make scripts successfully when creating pytests

- [x] I ran my code


fre/make/createCheckout.py
```diff
@@ -5,11 +5,10 @@
 import logging
 import sys
 import click
-from .gfdlfremake import varsfre, yamlfre, checkout, targetfre
 import fre.yamltools.combine_yamls as cy
+from .gfdlfremake import varsfre, yamlfre, checkout, targetfre
 
-@click.command()
-def checkout_create(yamlfile,platform,target,no_parallel_checkout,jobs,execute,verbose):
+def _checkout_create(yamlfile,platform,target,no_parallel_checkout,jobs,execute,verbose):
```
ilaflott, line 11: backwards! this should not have an underscore. Line 107 should

fre/make/createCompile.py
```diff
@@ -9,8 +9,7 @@
 from .gfdlfremake import varsfre, yamlfre, targetfre, buildBaremetal
 import fre.yamltools.combine_yamls as cy
 
-@click.command()
-def compile_create(yamlfile,platform,target,jobs,parallel,execute,verbose):
+def _compile_create(yamlfile,platform,target,jobs,parallel,execute,verbose):
```
ilaflott, line 12: see other comment

fre/make/createDocker.py
```diff
@@ -9,7 +9,7 @@
 import fre.yamltools.combine_yamls as cy
 
 @click.command()
-def dockerfile_create(yamlfile,platform,target,execute):
+def _dockerfile_create(yamlfile,platform,target,execute):
```
ilaflott, line 12: see above

ilaflott (changes requested):
underscore convention is backwards!

fre/make/createDocker.py
singhd789: ooooop, my bad. Let me fix that

singhd789:
> underscore convention is backwards!

@ilaflott Hmmm so when I reverse the underscore convention, I keep getting this: 
```
TypeError: Callback is not a command.
```

ilaflott:
> > underscore convention is backwards!
> 
> @ilaflott Hmmm so when I reverse the underscore convention, I keep getting this:
> 
> ```
> TypeError: Callback is not a command.
> ```

since it changes a module name, you need to change the name in other places where that old name is imported, is my guess.

this pattern is absolutely doable, most every other subtool currently has it

ilaflott:
> > > underscore convention is backwards!
> > 
> > 
> > @ilaflott Hmmm so when I reverse the underscore convention, I keep getting this:
> > ```
> > TypeError: Callback is not a command.
> > ```
> 
> since it changes a module name, you need to change the name in other places where that old name is imported, is my guess.
> 
> this pattern is absolutely doable, most every other subtool currently has it

let me note that `fre pp checkout` also has the pattern backwards

rem1776 (approved):

singhd789:
> let me note that `fre pp checkout` also has the pattern backwards... i.e. i'll get on others case about it!

I think I may have done it with ` fre pp configure-yaml ...` as well. 😅

ilaflott (approved):

PR #233, opened 2024-10-30
 Fix making checkout script executable before running

singhd789:
- want from users to be able to run checkout script themselves as well
- this is a fix: chmod needed to be moved to just be in `createCheckout`

- [x] I ran my code


ilaflott:
i think we just had a merge to address this?

can i suggest this problem have a corresponding unit test?

singhd789:
> i think we just had a merge to address this?
> 
> can i suggest this problem have a corresponding unit test?

@ilaflott Yes, the tests are being worked on. It was a very slight confusion on my part tbh. There were multiple places where `chmod` was used. When I initially tested, it worked, but that was because it worked for everything except when the container build steps are implemented. (I forgot to test one step).

rem1776 (approved):
Thanks! tested on gaea and it fixed any errors i was seeing with the container build

singhd789:
@ilaflott @rem1776 thanks! Writing up a little unit test to push here

Issue #226, opened 2024-10-28
fre make run-fremake test


singhd789:
**Is your feature request related to a problem? Please describe.**
fre make run-fremake needs an associated test.

**Describe the solution you'd like**
Tests could include 
- [ ] existence of model yaml, platform yaml, and compile yaml
- [ ] test for correct target options: Can only list one mutually exclusive target: prod, repro, debug (can list each separately with own -t but not together) --> Target can include openmp with those, but you cannot have both prod and repro listed in one target click entry, for example (not allowed: -t prod-repro)
- [ ] failure for nonexistent platform or target
- [ ] failure if pointing to nonexistent yaml
- [ ] successful creation of checkout script (container - checkout script would be in `tmp` location, non-container - checkout script would be in `modelRoot` location) 
- [ ] successful running of checkout script
- [ ] successful creation of makefile
- [ ] successful creation of compile script
- [ ] successful running of compile script
- [ ] successful creation of dockerfile (if `-npc` and container platform used)
- [ ] successful container build

PR #245, opened 2024-11-01
update CI image and add mkmf for fre make testing

### LINKED ISSUE #229: mkmf not included in installation
**Describe the bug**
mkmf is not included in the installation,  fre make will fail when trying to use it

**To Reproduce**
Install fre-cli and attempt to run fre make, it will fail because it cannot find mkmf in the path

**Expected behavior**
fre-cli should include mkmf (i.e all of its dependencies) 

**Additional context**
You can work around this by cloning mkmf and adding it to your path, or loading fre/bronx-22 if on gaea.

rem1776:
## Describe your changes
This PR updates the existing CI workflow that runs pytest to support compilation of models using fre make, as well as container creation.

- Changes the CI image to a newly created one with gcc, mpich and anything else needed to test fre make capabilities more thoroughly. This replaces the 'stock' image we were using from github and has miniconda 24.7.1 and gcc 14
- Adds mkmf as a git submodule
- Adds a new test file that checks the `run-fremake` subcommand compiles the null model successfully using an added CI platform
- Made changes to the workflow to run on the new image and cleaned up a few commands that don't seem necessary 

#226
Fixes #229 

- [x] I ran my code


ilaflott:
I'm not reviewing this yet, but i did want to bring your attention to my attempt to package `fms_yaml_tools` with `gfdl_msd_schemas` as a `git` submodule. If i'm not overlooking something, your approach will run into [this](https://github.com/NOAA-GFDL/fms_yaml_tools/pull/66#issuecomment-2419913056)

TBH, i don't think i exhausted all possibilities/workarounds, but from my googling at the time, the situation doesn't look fun. I would like to see you succeed with this, because then making `fms_yaml_tools` a proper conda package for `fre-cli` will be MUCH easier.

rem1776:
> I'm not reviewing this yet, but i did want to bring your attention to my attempt to package `fms_yaml_tools` with `gfdl_msd_schemas` as a `git` submodule. If i'm not overlooking something, your approach will run into [this](https://github.com/NOAA-GFDL/fms_yaml_tools/pull/66#issuecomment-2419913056)
> 
> TBH, i don't think i exhausted all possibilities/workarounds, but from my googling at the time, the situation doesn't look fun. I would like to see you succeed with this, because then making `fms_yaml_tools` a proper conda package for `fre-cli` will be MUCH easier. I.e. PLEASE prove me wrong about `conda build` / `git submodule`, i would really prefer to be wrong.

I haven't gotten that error in the CI testing or when I ran it in the container manually. Did you try starting off that CI with the recursive checkout option like i did for this one? I think pulling in the submodule after the `conda env create` might be part of the issue.

ilaflott:
> > I'm not reviewing this yet, but i did want to bring your attention to my attempt to package `fms_yaml_tools` with `gfdl_msd_schemas` as a `git` submodule. If i'm not overlooking something, your approach will run into [this](https://github.com/NOAA-GFDL/fms_yaml_tools/pull/66#issuecomment-2419913056)
> > TBH, i don't think i exhausted all possibilities/workarounds, but from my googling at the time, the situation doesn't look fun. I would like to see you succeed with this, because then making `fms_yaml_tools` a proper conda package for `fre-cli` will be MUCH easier. I.e. PLEASE prove me wrong about `conda build` / `git submodule`, i would really prefer to be wrong.
> 
> I haven't gotten that error in the CI testing or when I ran it in the container manually. Did you try starting off that CI with the recursive checkout option like i did for this one? I think pulling in the submodule after the `conda env create` might be part of the issue.

RE git submodules: [consider my anxieties salved!](https://github.com/NOAA-GFDL/fms_yaml_tools/pull/66#issuecomment-2458003152)

rem1776:
@ilaflott @singhd789 @kiihne-noaa This is ready now but wanted to bring up a potential side-effect of adding the run-fremake test.

I didn't know this initially but it looks like `conda build` actually runs all the pytests as part of building the package. I guess thats standard for conda, but in this case it could prevent you from doing a `conda build` on other systems since the platform used in the test is specific to the CI image. Without editing the `ci.gnu` platform (and sometimes the mkmf template) for whichever system you are on the compile will probably fail during the test and ruin the build.

I'm sure there's some way to exclude a specific test from pytest, so maybe it would be good to exclude compilation related tests during `conda build` for the sake of portability?

Any thoughts appreciated, I'm still a conda newb so lmk if i got something wrong.

singhd789:
> @ilaflott @singhd789 @kiihne-noaa This is ready now but wanted to bring up a potential side-effect of adding the run-fremake test.
> 
> I didn't know this initially but it looks like `conda build` actually runs all the pytests as part of building the package. I guess thats standard for conda, but in this case it could prevent you from doing a `conda build` on other systems since the platform used in the test is specific to the CI image. Without editing the `ci.gnu` platform (and sometimes the mkmf template) for whichever system you are on the compile will probably fail during the test and ruin the build.
> 
> I'm sure there's some way to exclude a specific test from pytest, so maybe it would be good to exclude compilation related tests during `conda build` for the sake of portability?
> 
> Any thoughts appreciated, I'm still a conda newb so lmk if i got something wrong.

@rem1776  Ooooh right, fre make is kind of specific to gaea for bare-metal builds. That's interesting. I didn't think about that, thank you for bringing it up. There is this: to skip certain tests in a test script, you can add `@pytest.mark.skip(reason='why we're skipping')` that you can add right before a test

rem1776:
@singhd789 Hmmm I'm not sure if we should add that cause then it'd be skipped every time and there'd be no point to having it.

It looks like there's an `--ignore` argument to the pytest command that can skip a directory, I think if we add it here it should skip it during the conda build but still include the test if you run pytest outside of the `conda build` command:

https://github.com/NOAA-GFDL/fre-cli/blob/20e943aa94b04ed2e50bfa37d19920eedba13c2b/meta.yaml#L66

ilaflott:
> @singhd789 Hmmm I'm not sure if we should add that cause then it'd be skipped every time and there'd be no point to having it.
> 
> It looks like there's an `--ignore` argument to the pytest command that can skip a directory, I think if we add it here it should skip it during the conda build but still include the test if you run pytest outside of the `conda build` command:
> 
> https://github.com/NOAA-GFDL/fre-cli/blob/20e943aa94b04ed2e50bfa37d19920eedba13c2b/meta.yaml#L66

we have a few situations where tests aren't going to run anywhere right away. IMO i'd rather have a skip on a test that works somewhere specific, than no test at all. there's a lot of iteration on these things happening across the board...

ilaflott:
> @ilaflott @singhd789 @kiihne-noaa This is ready now but wanted to bring up a potential side-effect of adding the run-fremake test.
> 
> I didn't know this initially but it looks like `conda build` actually runs all the pytests as part of building the package. 
this is actually a choice i made- if the tests can't run as part of `conda build`, i think we DO want the packaging step to fail. 


> I guess thats standard for conda, but in this case it could prevent you from doing a `conda build` on other systems since the platform used in the test is specific to the CI image. 
I think ideally we only publish the conda package from here as a `noplatform` build. the presence of compilers etc. is something i think the code should check for and handle/

> Without editing the `ci.gnu` platform (and sometimes the mkmf template) for whichever system you are on the compile will probably fail during the test and ruin the build.
> I'm sure there's some way to exclude a specific test from pytest, so maybe it would be good to exclude compilation related tests during `conda build` for the sake of portability?
i think this is fine for now

> Any thoughts appreciated, I'm still a conda newb so lmk if i got something wrong.
yer doing great work

meta.yaml
```diff
@@ -63,7 +63,7 @@ test:
     - fre.catalog
   commands:
     - pylint --max-args 6 -ry --ignored-modules netCDF4,cmor fre/ || echo "pylint returned non-zero exit code and will kill the workflow. guarding against this now."
-    - pip install GitPython && pytest --config-file=fre/pytest.ini --cov-config=fre/coveragerc --cov=fre fre/
+    - pip install GitPython && pytest --config-file=fre/pytest.ini --cov-config=fre/coveragerc --cov=fre --ignore=fre/make/tests/compilation fre/ # run pytest but ignore any tests that require compilation
```
ilaflott, line 66: mmmmm maybe we need a [`pytest.ini`](https://docs.pytest.org/en/stable/reference/customize.html) ?

fre/make/tests/compilation/test_fre_make_run_fremake.py
```diff
@@ -0,0 +1,13 @@
+''' test "fre make" calls '''
+
+from click.testing import CliRunner
+
+from fre import fre
+
+runner = CliRunner()
+
+## assumes you're running pytest from repo root dir
+def test_fre_make_run_fremake_null_model():
```
singhd789, line 10: I think if we want to test the importability of scripts and the functionality of this tool, it would be better to have this actually use the functions inside of `run_fremake`. So maybe instead of using `runner.invoke(...)`:

At the top:
```
from fre.make import runFremake 
```

Then testing the tool
```
runFremake.fremake_run(yamlfile,platform,target)
```
Where the `assert` could check for things like a checkout script, makefile, and compile script, and that they were successful

From the way @ilaflott described it, my understanding is that defining the runner is good for testing click usage / options which can be done in the `fre/tests/test_fre_make_cli.py` file

meta.yaml
rem1776: To ignore the tests? Or to add the directory?

The parent directory is in the current pytest.ini, which should cover this as well I think:
https://github.com/NOAA-GFDL/fre-cli/blob/20e943aa94b04ed2e50bfa37d19920eedba13c2b/fre/pytest.ini#L8

fre/make/tests/compilation/test_fre_make_run_fremake.py
ilaflott: yeah that is my general position- `CliRunner` based tests in `fre/tests`, `from fre.subtool import FOO` style tests in `fre/subtool/tests`

i'm also sort of waiting for someone to show me why that pattern maybe won't and/or can't last, if that is the case

meta.yaml
ilaflott: ah i'm wondering if we can use it to ignore, but maybe we could do an explicit allow-list too, the list wouldn't be too long... yet.

ilaflott: that being said this definitely works, so what i'm pointing out is really a nit for now, priority wise

fre/make/tests/compilation/test_fre_make_run_fremake.py
rem1776: Oh yeah I think we talked about that! I should have those changes somewhere(hopefully), sorry got a little sidetracked by the image/conda debugging. I'll get the actual test updated soon.

meta.yaml
rem1776: It looks like there's a way to add flags to that file, but if the ignore flag was added it would always be skipped by pytest, not just during conda builds.

fre/make/tests/compilation/test_fre_make_run_fremake.py
rem1776: Should be fixed now. Right now it just has the run-fremake compile and checks for the resulting executable, I'll add a lot more tests to the file once this CI stuff is all merged in.

.github/workflows/build_conda.yml
```diff
@@ -7,12 +7,15 @@ jobs:
   build:
     runs-on: ubuntu-latest
     container:
-      image: continuumio/miniconda3:latest
+      image: ghcr.io/noaa-gfdl/fre-cli:miniconda24.7.1_gcc14.2.0
```
singhd789, line 10: Clarifying question for myself, is this a version/image we'll have to update every once in a while?

.github/workflows/create_test_conda_env.yml
```diff
@@ -5,67 +5,54 @@ on: [push]
...
-          
-          # are we talking to the right python?
-          which python
-          python --version
-          $CONDA/envs/fre-cli/bin/python --version
-          
+          # add spack installed binaries to front of path so that
+          # conda's netcdf/hdf5 installs don't break compilation tests
+          export path_save=$PATH
+          export PATH="/opt/views/view/bin:$PATH"
```
singhd789, line 37: Tom just walked me through views the other day, cool to see!

.github/workflows/create_test_conda_env.yml
```diff
@@ -5,67 +5,54 @@ on: [push]
...
+
       - name: Run pylint in fre-cli environment
         run: |
-          # try to make sure the right things are in GITHUB_PATH
-          echo $CONDA/envs/fre-cli/bin >> $GITHUB_PATH
-
-          # are we talking to the right python?
-          which python
-          python --version
-          $CONDA/envs/fre-cli/bin/python --version
```
singhd789, line 61: Another clarifying question, for @rem1776 or @ilaflott or whoever knows, why were we checking for the python version to begin with?

.github/workflows/publish_conda.yml
```diff
@@ -7,12 +7,15 @@ jobs:
...
     runs-on: ubuntu-latest
     container:
-      image: continuumio/miniconda3:latest
+      image: ghcr.io/noaa-gfdl/fre-cli:miniconda24.7.1_gcc14.2.0
     steps:
       - name: Checkout Files
         uses: actions/checkout@v4
         with:
           submodules: 'recursive'
+      - name: Add mkmf to PATH
```
singhd789, line 16: Oooo, do we need to add this for submodules? The submodules wouldn't just be built with the conda package with the `submodules: 'recursive'`?

singhd789 (approved):

ilaflott (approved):

.github/workflows/build_conda.yml
rem1776: I don't think anything will break anytime soon if we used this image for a while, but it'd be good to get a new image with newer software every year or so, that's what I've typically done with other CI images. Anytime you rebuild the image it'll pull in the latest versions, so you wouldn't have to make any changes to the dockerfile.

ilaflott: out of curiosity and my own notekeeping, where can one find the dockerfile for the uploaded container image in question?

rem1776: It's in the hpc-me repo:

https://gitlab.gfdl.noaa.gov/fre/HPC-ME/-/blob/main/ci/fre-cli-gnu/Dockerfile?ref_type=heads

singhd789:
The only thing we might have to discuss is the output location for fre make tests. The bare-metal runs all have the same output location of `HOME/fremake_canopy/test` and I have a feeling tests might conflict or fail if files are getting overwritten on removed as the tests run. Something we should explore

ilaflott:
> The only thing we might have to discuss is the output location for fre make tests. The bare-metal runs all have the same output location of `HOME/fremake_canopy/test` and I have a feeling tests might conflict or fail if files are getting overwritten on removed as the tests run. Something we should explore

is the output location hard coded? it should, in theory, be something that can be set elsewhere and fed through as an e.g. an argument

PR #246, opened 2024-11-04
update function names for run_fremake subcommand

rem1776:
## Describe your changes
Some changes needed to get the run-fremake subcommand working. Just switches around the names for `run_fremake` and adds a chmod.

- [x] I ran my code

### DIFF
--- a/fre/make/gfdlfremake/checkout.py
+++ b/fre/make/gfdlfremake/checkout.py
@@ -115,6 +115,7 @@ class checkout():
         Param:
             - self The checkout script object
         """
+        os.chmod(self.src+"/"+self.fname, 0o744)
         try:
             subprocess.run(args=[self.src+"/"+self.fname], check=True)
         except:
--- a/fre/make/gfdlfremake/checkout.py
+++ b/fre/make/gfdlfremake/checkout.py
@@ -111,11 +111,10 @@ class checkout():
 ## TODO: batch script building
     def run (self):
         """
-        Brief: Changes the permission on the checkout script and runs it
+        Brief: Runs the checkout script
         Param:
             - self The checkout script object
         """
-        os.chmod(self.src+"/"+self.fname, 0o744)
         try:
             subprocess.run(args=[self.src+"/"+self.fname], check=True)
         except:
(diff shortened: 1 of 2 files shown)


fre/make/gfdlfremake/checkout.py
```diff
@@ -115,6 +115,7 @@ def run (self):
         Param:
             - self The checkout script object
         """
+        os.chmod(self.src+"/"+self.fname, 0o744)
```
singhd789, line 118: I don't think this would be needed here. The checkout script being made executable is now handled in the `createCheckout` tool: https://github.com/NOAA-GFDL/fre-cli/blob/c82680493b2a650b36011f0614c22b62117b31c6/fre/make/createCheckout.py#L78

fre/make/runFremake.py
```diff
@@ -200,12 +200,12 @@ def _fremake_run(yamlfile,platform,target,parallel,jobs,no_parallel_checkout,ver
...
 
 @click.command()
-def fremake_run(yamlfile,platform,target,parallel,jobs,no_parallel_checkout,verbose):
+def _fremake_run(yamlfile,platform,target,parallel,jobs,no_parallel_checkout,verbose):
     '''
     Decorator for calling _fremake_run - allows the decorated version
     of the function to be separate from the undecorated version
     '''
-    return _fremake_run(yamlfile,platform,target,parallel,jobs,no_parallel_checkout,verbose)
+    return fremake_run(yamlfile,platform,target,parallel,jobs,no_parallel_checkout,verbose)
```
singhd789, line 208: These are supposed to be switched, great catch! Thank you

singhd789 (changes requested):

fre/make/gfdlfremake/checkout.py
singhd789: Users want the checkout script to be made executable for the bare-metal build before actually being able to run it. I made that change in the `createChecout.py` tool and forgot to change it for `run_fremake` as well. I would put this `chmod` line here: https://github.com/NOAA-GFDL/fre-cli/blob/34eb37e2130ddf4703f343f0ccc9bcc38d2102b8/fre/make/runFremake.py#L92

Right after the checkout script gets created.

rem1776: This should be fixed now with commit ee4b5b5. Will run-fremake not always execute everything? I think thats what it does currently.

singhd789 (approved):

PR #253, opened 2024-11-05
187.schema submodule

singhd789:

- [x] I ran my code


meta.yaml
```diff
@@ -5,10 +5,10 @@ package:
   version: '{{ environ.get("GIT_DESCRIBE_TAG", data.get("version")) }}'
 
 source:
-  path: .
+#  path: .
 # ideally we want this git_url path, but it messes with conda publish
 # where it builds only the sourced url and not local/branch changes
-# git_url: https://github.com/NOAA-GFDL/fre-cli
+  git_url: https://github.com/NOAA-GFDL/fre-cli.git
```
ilaflott, line 11: wait are you telling me the problem was the lack of `.git` ??

fre/make/gfdlfremake/yamlfre.py
```diff
@@ -176,8 +177,8 @@ def __init__(self,combinedyaml,v):
         #self.freyaml.update(self.platformsyaml)
 
         ## VALIDATION OF COMBINED YAML FOR COMPILATION
-        fremake_package_dir = os.path.dirname(os.path.abspath(__file__))
-        schema_path = os.path.join(fremake_package_dir, 'schema.json')
+        fremake_package_dir = Path(__file__).resolve().parents[2]
+        schema_path = os.path.join(fremake_package_dir, 'gfdl_msd_schemas', 'FRE', 'fre_make.json')
```
ilaflott, line 181: not sure the `parents` thing is worth it here, `fre/make` is pretty unlikely to ever change in location.

fre/yamltools/tests/test_combine_yamls.py
```diff
@@ -81,8 +81,8 @@ def test_combined_compileyaml_validation():
     Validate the combined compile yaml
     """
     combined_yamlfile =f"{COMP_OUT_DIR}/combined-{COMP_EXPERIMENT}.yaml"
-    schema_file = os.path.join(f"{IN_DIR}","compile_yamls","schema.json")
-
+    schema_file = os.path.join(Path(TEST_DIR).resolve().parents[1], "gfdl_msd_schemas", "FRE", "fre_make.json")
```
ilaflott, line 84: similar comment RE `.parents` and, while we may move `gfdl_msd_schemas` idk who knows, might be clearer to just have the spot hard coded. will probably break more clearly than a `os.path` function call

fre/yamltools/tests/test_combine_yamls.py
```diff
@@ -134,7 +134,7 @@ def test_combined_compileyaml_validatefail():
 
     # Validate against schema; should fail
     wrong_combined = Path(f"{COMP_OUT_DIR}/combined-am5-wrong_datatype.yaml")
-    schema_file = os.path.join(f"{IN_DIR}","compile_yamls","schema.json")
+    schema_file = os.path.join(Path(TEST_DIR).resolve().parents[1], "gfdl_msd_schemas", "FRE", "fre_make.json")
```
ilaflott, line 137: see above

fre/make/gfdlfremake/schema.json
ilaflott, line 1: bless for you deleting things

fre/gfdl_msd_schemas
ilaflott, line 1: the comment i'm going to leave here is likely for a future PR, but is this where we want this? if `fms_yaml_tools` simultaneously lists `gfdl_msd_schemas` in it's submodule list too, and we link up `fre/yamltools` to `fms_yaml_tools` and have `fms_yaml_tools` listed as a conda package req, then might having two `fms_yaml_tools` floating around the conda packages be a recipe for trouble? 

thoughts....

ilaflott (approved):
I'm going to approve but with the nit of- the `.parents` manipulation in the test scripts look more trouble than it's worth. I'll leave it up to @singhd789 to make the final call on that, though

singhd789:
> I'm going to approve but with the nit of- the `.parents` manipulation in the test scripts look more trouble than it's worth. I'll leave it up to @singhd789 to make the final call on that, though

Hmm, you might be right. I'll look into it a little more before merging.

meta.yaml
singhd789: It might've been. I was following an example I saw somewhere and they included the `.git`

fre/make/gfdlfremake/yamlfre.py
singhd789: I just saw it as another easy way to define absolute paths / find the directory names ... I'll look into if it can introduce issues

fre/gfdl_msd_schemas
singhd789: Wait, sorry, why would we want to have two `fms_yaml_tools` floating around?

fre/yamltools/tests/test_combine_yamls.py
singhd789: I see what you mean, especially for tests. Might be better to have that path laid out explicitly

fre/make/gfdlfremake/yamlfre.py
singhd789: I think I'll keep this one here. It finds the right schema location in reference to this file itself, imo, in a simple way

fre/gfdl_msd_schemas
ilaflott: if `gfdl_msd_schemas` is a submodule of `fre-cli` and `fms_yaml_tools`, and `fms_yaml_tools` were a conda package dependency of `fre-cli`, then one would have both `fre/gfdl_msd_schemas` and `fms_yaml_tools/schema/gfdl_msd_schemas` floating around in the environment's `site-packages` directory. At worst, one may worry about an ambiguity arising between the two, or clashing information. At best, it's redundant and not satisfying.

It's not something that needs fixing now- it's a weak spot to jot down for future improvement.

meta.yaml
ilaflott: OK, if we end up with a mysterious problem at the `publish_conda` step, we'll know @bcc2761's old comment still applies.

PR #259, opened 2024-11-07
Documentation

ceblanton:

- [ ] I ran my code


ilaflott:
the success/fail between the two checks is awkward... but i see some pull's happening in the history, so maybe one is being re-run when the other isn't or smth. 

i don't see anything that explains the failure, having to do with the work here or otherwise... hopefully that succeeds on the next run.

ceblanton:
Even though I'm still working on these, I think we should merge them in. As we are all updating docs today, we'll see more merge conflicts unless we merge more frequencly.

ilaflott:
> Even though I'm still working on these, I think we should merge them in. As we are all updating docs today, we'll see more merge conflicts unless we merge more frequently.

still in draft mode! no one can merge it in, even with approval, if the PR is in draft mode.

ilaflott (approved):
assuming the check passes, i approve.

PR #260, opened 2024-11-08
Delete some things

ilaflott:
quick effort- unused things as far as i know.


ilaflott:
something kinda funny about the way `coveragerc` is or is-not getting picked up by `pytest-cov`... in any case, ready to go.

fre/pytest.ini
singhd789, line 1: I don't know too much about pytest.ini and coveragerc files. In a quick search, the pytest.ini file is a config file that can be used to specify any settings about pytests. Can the coveragerc file be combined in here since they are both about the tests?

singhd789 (approved):
THANK YOU FOR THIS MUCH NEEDED CLEAN UP

fre/pytest.ini
ilaflott: not quite, the coverage tool `pytest-cov`, is actually another third party tool. so unless that kind of config appending is supported explicitly, i assume it can't be done. 

i feel your feelings here though

PR #271, opened 2024-11-20
add --execute option to run-fremake

rem1776:
## Describe your changes
Changes run-fremake to work more like bronx's fremake and require a `-e`/`--execute` flag to run build scripts.

If the `--execute` option is given, it'll work as it does currently without the flag and run the compile script/docker build command after it is created. If its not given, it'll print out a compile script path for the user to run.

Since container builds don't have a build script currently this also adds code to write out a short script with the container commands to build and transfer the format.

- [x] I ran my code


fre/make/gfdlfremake/buildDocker.py
```diff
@@ -187,14 +187,35 @@ def writeRunscript(self,RUNenv,containerRun,runOnDisk):
...
-            - containerRun : The container platform used with `exec` to 
+            - containerRun : The container platform used with `exec` to
                              run the container; apptainer or singularity used
         """
         os.system(containerBuild+" build -f Dockerfile -t "+self.e+":"+self.target.gettargetName())
         os.system("rm -f "+self.e+".tar "+self.e+".sif")
         os.system(containerBuild+" save -o "+self.e+"-"+self.target.gettargetName()+".tar localhost/"+self.e+":"+self.target.gettargetName())
         os.system(containerRun+" build --disable-cache "+self.e+"-"+self.target.gettargetName()+".sif docker-archive://"+self.e+"-"+self.target.gettargetName()+".tar")
+
+    def createBuildScript(self,containerBuild,containerRun):
```
singhd789, line 202: Great idea, thanks for this! Takes a little bit more of the magic of how to build the actual container away. 🎉 Maybe we should actually make the creation of this script a default behavior in `def build`. (basically replacing `def build` with `def createBuildScript`) 
Then in `runFremake.py`, instead of 
```
if execute:
    dockerBuild.build(containerBuild,containerRun)
else:
    dockerBuild.createBuildScript(containerBuild, containerRun)
    print("Container build script created at "+dockerBuild.userScriptPath+"\n\n")
```
It would be:
```
#create the container build script
dockerBuild.createBuildScript(containerBuild, containerRun)

#print statement 
(ADD print statement somewhere about where the script is located)

if execute:
    subprocess.run(args=[dockerBuild.userScriptPath], check=True)
else:
    print("Container build script created at "+dockerBuild.userScriptPath+"\n\n")
```

What do you think? Or do you think it might be better to have one less script being made if it doesn't need to be (i.e. if execute is passed, no need to make the build script)

fre/make/runFremake.py
```diff
@@ -187,26 +190,32 @@ def fremake_run(yamlfile,platform,target,parallel,jobs,no_parallel_checkout,verb
 
                 dockerBuild.writeRunscript(RUNenv,containerRun,tmpDir+"/execrunscript.sh")
 
-                ## Run the dockerfile; build the container
-                dockerBuild.build(containerBuild,containerRun)
+                ## Run the dockerfile and create the container if execute option is given
+                ## otherwise create a build script and print out its path
+                if execute:
```
singhd789, line 195: These steps would also need to be added in `createDocker.py` as well

singhd789: Oh wait, to clarify, `createDocker.py` already has the execute (run) or not options. If execute, it'll created the dockerfile and build. If not, it'll just create the dockerfile. So it would need the ability to create the buildScript.

singhd789: But maybe that can be done in another PR since this is mainly for run-fremake

fre/make/gfdlfremake/buildDocker.py
rem1776: Yeah I think always making the script is a good idea just cause it'll be a little more transparent and easier to debug build issues. I'll add those changes in!

fre/make/runFremake.py
rem1776: Sounds good, I can put those changes in another PR.

singhd789 (approved):

PR #273, opened 2024-11-22
Add analysis tools

menzel-gfdl:
## Describe your changes
Add some analysis tools to:
- run analysis scripts
- add experiments to dora (will require changes to dora to actually work)
- publish analysis figures to dora (will require changes to  dora to actually work)

- [ ] I ran my code


meta.yaml
```diff
@@ -30,6 +30,7 @@ requirements:
     - jsonschema
     - noaa-gfdl::catalogbuilder
     - noaa-gfdl::fre-nctools
+    - noaa-gfdl::analysis_scripts
```
ilaflott, line 33: this needs to be in `environment.yml` as well, for tests/building in `.github/workflows/create_test_conda_env.yaml`

ilaflott (changes requested):
Well aware this is still in draft mode, but this is not far off from being able to be merged at all. Excellent work!

This needs the new `noaa-gfdl::analysis_scripts` package req added to `environment.yml`. If that module is indended to be pure-python package, we should add it to `setup.py` as well. 

Without this, a currently doc'd development approach involving `conda env create -f environment.yml` breaks, and thats why `create_test_conda_env.yaml` is likely failing. 

After doing that and making sure the pipeline passes, merge the latest 100 or commits to `main` into this branch to make sure nothing clashes. I doubt any clashes will come up... but no need to speculate, we have a pipeline!

Thrilled to approve this when its ready

meta.yaml
menzel-gfdl: @ilaflott Thank you, I have added that to the `environment.yml`.

menzel-gfdl: Oh I see, `fre/tests` (not `fre/test`), my bad

fre/analysis/freanalysis.py
```diff
@@ -0,0 +1,63 @@
+import click
+
+from .subtools import install_analysis_package, run_analysis, uninstall_analysis_package
```
ilaflott, line 3: so these things will need separate, protected-entry points for `click` decorators in `fre/freanalysis/subtools.py` if module-driven usage is at all a design here (it absolutely should IMO).

In anycase, if you decorate the functions in `subtools` now as they are, `click` will block imports of those functions elsewhere.

fre/analysis/freanalysis.py
```diff
@@ -0,0 +1,63 @@
...
+              help="Path to the output directory.")
+@click.option("--output-yaml", type=str, required=True, help="Path to the output yaml.")
+@click.option("--experiment-yaml", type=str, required=True, help="Path to the experiment yaml.")
+@click.option("--library-directory", type=str, required=False,
+              help="Path to a custom lib directory.")
+@click.pass_context
+def run(context, name, catalog, output_directory, output_yaml, experiment_yaml,
+        library_directory):
+    """Runs the analysis script and writes the paths to the created figures to a yaml file."""
+    run_analysis(name, catalog, output_directory, output_yaml, experiment_yaml,
```
ilaflott, line 48: so, for example, here, i'm going to suggest this to be `_run_analysis`

fre/analysis/subtools.py
```diff
@@ -0,0 +1,103 @@
...
+            chdir(tmp / "scripts" / "core" / "figure_tools")
+            install(".", library_directory)
+            chdir(tmp / "scripts" / "user-analysis-scripts" / name)
+            install(".", library_directory)
+            chdir(go_back_here)
+    else:
+        install(f"{url}@main", library_directory)
+
+
+def run_analysis(name, catalog, output_directory, output_yaml, experiment_yaml,
```
ilaflott, line 55: so, if we didn't care about simultaneously having both `fre analysis FOO` calls and `from fre.freanalysis.subtools import FOO`, the solution would simply add a `@click.command()` to L54 above here.

But to preseve the import, define + dectorate another function with `@click.command()`, call it `_run_analysis`. all this function will do is talk to click, and pass arguments to `run_analysis`.

@singhd789 recently tidied all of this up [in `fre/pp`](https://github.com/NOAA-GFDL/fre-cli/blob/0ad7e5cb3dda25ab82d77c0c72624380819a15de/fre/pp/frepp.py) recently, making it a pretty clean one to follow.

fre/analysis/tests/test_subtools.py
```diff
@@ -0,0 +1,57 @@
+from pathlib import Path
+import pytest
+from tempfile import TemporaryDirectory
+
+from analysis_scripts import find_plugins, available_plugins, UnknownPluginError
+from fre.analysis.subtools import install_analysis_package, run_analysis
+
+
+def make_experiment_yaml(path, name, whitespace="  "):
```
ilaflott, line 9: i haven't dug into every last thing about `pytest` yet, but I would double check to make sure this function gets used the way you think its getting used and isn't being collected as a test.

there's definitely room in `pytest`'s framework for canon-izing these sorts of set-up routines. i'll get there someday if someone else doesn't first, but for now, i superstitiously start every `pytest` with `test_`

fre/fre.py
```diff
@@ -21,7 +21,8 @@
                        "yamltools": ".yamltools.freyamltools.yamltools_cli",
                        "make": ".make.fremake.make_cli",
                        "app": ".app.freapp.app_cli",
-                       "cmor": ".cmor.frecmor.cmor_cli"                      },
+                       "cmor": ".cmor.frecmor.cmor_cli",
+                       "analysis": ".analysis.freanalysis.analysis_cli"},
```
ilaflott, line 25: another thing i doubt is the source of the current failures, but `python` is never a big fan of tabs, and this looks like a spurious tab got in here somehow!

ilaflott (changes requested):
This might be a lesson in why one never says, "it's almost done!" Apologies for bringing a pox-upon-the-house!

I'm pretty sure this is getting caught up on some of `click` and `python`'s most classic pit falls. I think I caught most of them.

fre/fre.py
ilaflott: wait- this might have been a text-on-my-screen rendering thing. possibly not an issue, but also not impossible github doesnt render tabs everywhere with the same conversion rule.

fre/analysis/freanalysis.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@analysis_cli.command()
+@click.option("--name", type=str, required=True, help="Name of package to uninstall.")
+@click.option("--library-directory", type=str, required=False,
+              help="Path to a custom lib directory.")
+@click.pass_context
+def uninstall(context, name, library_directory):
+    """Uninstall an analysis package."""
+    uninstall_analysis_package(name, library_directory)
```
ilaflott, line 59: this needs to be a `context.forward(uninstall_analysis_package)`, and the lines similar to it

meta.yaml
menzel-gfdl: I'm not sure how to add it to the `setup.py`, since it is in a subdirectory of a github repo.  I'll try to do some research.

fre/tests/test_fre_analysis_cli.py
```diff
@@ -0,0 +1,102 @@
...
+from click.testing import CliRunner
+
+from fre import fre
+
+
+runner = CliRunner()
+
+
+def test_cli_fre_analysis():
+    """Most basic invocation of `fre analysis`."""
```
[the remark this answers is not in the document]
menzel-gfdl: Should I change them to single quotes?

fre/tests/test_fre_analysis_cli.py
```diff
@@ -0,0 +1,102 @@
+"""Test "fre analysis" tools."""
```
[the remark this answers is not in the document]
menzel-gfdl: Change these to single quotes too?

fre/tests/test_fre_analysis_cli.py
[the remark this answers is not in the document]
ilaflott: i think most of these doc strings in the other test scripts are just the cli-equivalent call, `fre analysis`. should be sufficient. One thing to keep in mind i suppose is sometimes auto-doc packages have to parse the doc-strings too. That pushes my opinion to, "fewer escaping-curveballs in doc-strings where possible"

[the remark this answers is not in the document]
ilaflott: see above!

meta.yaml
ilaflott: ah, we dont have `catalogbuilder` there either. must be something about the way the package's metadata gets set, but I don't know exactly what. Not important- doesn't affect functionality. `conda install` picks up the slack just fine, presumably because the equivalent info is in `meta.yaml`

PR #277, opened 2024-11-26
add container buildscript creation for `createDockerfile`

rem1776:
## Describe your changes
Updates the `createDockerfile` fremake subcommand to create a short container build script instead of running commands directly.

I also added another level of indents so it'll only create the build script/run if the `if iscontainer` conditional is true like the commands above it. I think it needs the commands above it for the build script to work properly, but I don't really know how multiplatform builds work so let me know if it should be moved back.

This needs #271 in first, which adds the same functionality to run-fremake with container builds.

- [ ] I ran my code

### DIFF
--- a/fre/make/createDocker.py
+++ b/fre/make/createDocker.py
@@ -72,10 +72,13 @@ def dockerfile_create(yamlfile,platform,target,execute):
                 click.echo("\ntmpDir created in " + currDir + "/tmp")
                 click.echo("Dockerfile created in " + currDir +"\n")
 
-            if run:
-                dockerBuild.build(containerBuild, containerRun)
-            else:
-                sys.exit()
+                # create build script for container
+                dockerBuild.createBuildScript(containerBuild, containerRun)
+                print("Container build script created at "+dockerBuild.userScriptPath+"\n\n")
+
+                # run the script if option is given
+                if run:
+                    subprocess.run(args=[dockerBuild.userScriptPath], check=True)
 
 @click.command()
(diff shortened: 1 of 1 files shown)


fre/make/createDocker.py
```diff
@@ -72,10 +72,13 @@ def dockerfile_create(yamlfile,platform,target,execute):
                 click.echo("\ntmpDir created in " + currDir + "/tmp")
                 click.echo("Dockerfile created in " + currDir +"\n")
 
-            if run:
-                dockerBuild.build(containerBuild, containerRun)
-            else:
-                sys.exit()
+                # create build script for container
+                dockerBuild.createBuildScript(containerBuild, containerRun)
```
singhd789, line 76: Hmm ok, I thought this looked ok but for some reason I'm seeing this when testing:
```
AttributeError: 'container' object has no attribute 'createBuildScript'
```

singhd789: WAIT, I think I forgot to install the recent changes in my conda environment

fre/make/createDocker.py
```diff
@@ -72,10 +72,13 @@ def dockerfile_create(yamlfile,platform,target,execute):
...
+                print("Container build script created at "+dockerBuild.userScriptPath+"\n\n")
+
+                # run the script if option is given
+                if run:
+                    subprocess.run(args=[dockerBuild.userScriptPath], check=True)
```
singhd789, line 81: Just need to import subprocess on the top!

singhd789 (approved):

PR #290, opened 2024-12-11
Null tests.142

kiihne-noaa:
## Describe your changes
Added tests in fre/make/null_example for checkout

creating this so Dana can help with debugging of issues with paths

@singhd789


fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,45 @@
+#tests for the create-checkout step of fre-make, for null_model.yaml
+import os
+from fre import fre
```
singhd789, line 3: Can remove this, not needed

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,45 @@
...
+from pathlib import Path
+from fre.pp import configure_script_yaml as csy
+from click.testing import CliRunner
```
singhd789, line 6: Can remove these too, not needed. We won't be using `runner.invoke` for these tool tests here

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,45 @@
...
+import subprocess
+from fre.make import createCheckout
```
singhd789, line 8: This needs to be updated: `createCheckout` --> `create_checkout_script`

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Set example yaml paths, input directory
+test_dir = Path("fre/make/tests/null_example")
+yamlfile = Path(f"{test_dir}/null_model.yaml")
+
+#set platform and target
+platform = "ncrc5.intel"
+target = "debug"
```
singhd789, line 16: To keep consistency with other tests, make these variable names capitalized.

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,45 @@
...
+from fre import fre
+from pathlib import Path
+from fre.pp import configure_script_yaml as csy
+from click.testing import CliRunner
+import subprocess
+from fre.make import createCheckout
+
+# Set example yaml paths, input directory
+test_dir = Path("fre/make/tests/null_example")
+yamlfile = Path(f"{test_dir}/null_model.yaml")
```
singhd789, line 12: I'd recommend keeping all the tests together in fre/make/tests for now. We can move it there and make this 

`TEST_DIR = Path("fre/make/tests")`
`YAMLFILE = Path(f"{TEST_DIR}/null_example/null_model.yaml")`

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,45 @@
...
+platform = "ncrc5.intel"
+target = "debug"
+
+#set output directory
+#out_dir = Path(f"fre/make/tests/null_example/fre_make_out")
+#Path(out_dir).mkdir(parents=True,exist_ok=True)
+
+# Set home for ~/cylc-src location in script
+#os.environ["HOME"]=str(Path(f"{out_dir}"))
+HOME_DIR = os.environ["HOME"]
```
singhd789, line 24: So through doing this, we've found it definitely effects other tests that use `HOME`. 

1) Let's move re-setting `HOME` to inside each test that needs it AND make sure to save the default `HOME`  location first. We can look at how Ryan maneuvered it:
define home: https://github.com/NOAA-GFDL/fre-cli/blob/3c9cacb99495a29032ef0f81e0d375963134af01/fre/make/tests/test_create_makefile.py#L54
set home: https://github.com/NOAA-GFDL/fre-cli/blob/3c9cacb99495a29032ef0f81e0d375963134af01/fre/make/tests/test_create_makefile.py#L55
reset home: https://github.com/NOAA-GFDL/fre-cli/blob/3c9cacb99495a29032ef0f81e0d375963134af01/fre/make/tests/test_create_makefile.py#L64

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,45 @@
...
+test_dir = Path("fre/make/tests/null_example")
+yamlfile = Path(f"{test_dir}/null_model.yaml")
+
+#set platform and target
+platform = "ncrc5.intel"
+target = "debug"
+
+#set output directory
+#out_dir = Path(f"fre/make/tests/null_example/fre_make_out")
+#Path(out_dir).mkdir(parents=True,exist_ok=True)
```
singhd789, line 20: In my opinion, I feel like it would be nice it each test had output that was easy to find. Here, for the checkout, maybe we can set OUT to be:
```
OUT = f"{TEST_DIR}/checkout_out"
```
Where test_dir would be what I commented above. Also add in the out location to the `.gitignore` file

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Set home for ~/cylc-src location in script
+#os.environ["HOME"]=str(Path(f"{out_dir}"))
+HOME_DIR = os.environ["HOME"]
+#run checkout command
+runner = CliRunner()
```
singhd789, line 26: remove

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,45 @@
...
+
+def test_checkout_script_exists():
+    """
+    Make sure checkout file exists
+    """
+    result = runner.invoke(fre.fre, args=["make","create-checkout","-y",yamlfile,"-p",platform,"-t",target])
```
singhd789, line 32: Using runner.invoke is good for testing the cli for fre. For actually script functionality testing here, we're going to call the actual functions in the script. 

Using `from fre.make import create_checkout_script` from the top:
1) we'll set/reset home as I mentioned above
2) we'll set what would be click options passed by the user for the fre make create-checkout command
3) we'll invoke the main checkout function 
4) check for the existence of the checkout script as a good starting test

You have line 33 that just needs to be edited a little. Try this:
`create_checkout_script.checkout_create(variables set right before)`

fre/make/tests/test_basic_null_yaml.py
singhd789, line 1: Is this different than the `test_basic_null_yaml.py` file in `fre/make/tests/null_example`?

singhd789 (changes requested):

fre/make/tests/test_basic_null_yaml.py
```diff
@@ -0,0 +1,27 @@
...
+platform = "gfdl.ncrc5-intel22-classic"
+target = "prod-openmp"
+
+# Set example yaml paths, input directory
+CWD = Path.cwd()
+test_dir = Path("fre/make/tests")
+test_yaml = Path(f"null_example/null_model.yaml")
+
+# Set home for ~/cylc-src location in script
+os.environ["HOME"]=str(Path(f"{CWD}/{test_dir}/configure_yaml_out"))
```
singhd789, line 16: this is probably why we keep seeing some failures

fre/make/tests/test_basic_null_yaml.py
singhd789, line 1: Is there a difference between this file and `fre/make/tests/null_example/test_basic_null_yaml.py`? I think we can get rid of the one in `null_example/`

fre/make/tests/null_example/test_checkout_null_yaml.py
singhd789, line 1: 2 things:
1) Can we put this with the other tests too? (Just in `fre/make/tests` - back one directory) 
2) Can we rename this `test_create_checkout.py`? I think the trend is to name the test file after the tool it's testing.

ilaflott:
for the `rm -rf`, [instead of subprocess](https://github.com/NOAA-GFDL/fre-cli/commit/baa9293d2f01e796db658a248227a082810d8e5c#diff-d6bf01611bc28cfd2ef62119818b16efc1eb48f8f7aa5b6754b98f22aeb6f5f2R40), consider `shutil.rmtree`

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,41 @@
...
+# Set home for ~/cylc-src location in script
+#run checkout command
+OUT = f"{TEST_DIR}/configure_yaml_out"
+os.environ["TEST_BUILD_DIR"] = OUT
+
+def test_checkout_script_exists():
+    """
+    Make sure checkout file exists
+    """
+    subprocess.run(["rm","-rf",f"{OUT}/fremake_canopy/test/null_model_full/src/checkout.sh"])
```
ilaflott, line 25: the `fremake_canopy/` part seems old/out-of-date. double check the directory structure being created locally and adjust what path you're attempting to remove and then check for in the `assert` below

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,41 @@
...
+    create_checkout_script.checkout_create(YAMLFILE,PLATFORM,TARGET,False,False, False, False)
+    #assert result.exit_code == 0
+    assert Path(f"{OUT}/fremake_canopy/test/null_model_full/src/checkout.sh").exists()
+
+def test_checkout_verbose():
+    """
+    check if --verbose option works
+    """
+    create_checkout_script.checkout_create(YAMLFILE,PLATFORM,TARGET,False,False, False,True)
```
ilaflott, line 34: @singhd789 @thomas-robinson perhaps `create_checkout_script` should `return` the full path to `checkout.sh`? instead of just printing it to screen?

singhd789: Well, the whole path to the checkout script is printed (and it's made executable) - that way the user knows where it is and could also run it themselves if they want. What would be another benefit of returning it as well? Just curious

fre/make/tests/null_example/test_checkout_null_yaml.py
singhd789: Hm, yeah the `fremake_canopy`  bit is actually a part of a path set in the `platforms.yaml`. Good point, it should be updated, but I don't thiiiiink that specific part is the problem. I still agree, always double check the path/directory structure for sure though!

fre/make/create_checkout_script.py
```diff
@@ -18,7 +18,9 @@ def checkout_create(yamlfile, platform, target, no_parallel_checkout, jobs, exec
     run = execute
     jobs = str(jobs)
     pcheck = no_parallel_checkout
-
+    
+    if jobs == 'False' and execute == True:
+        sys.exit('jobs must be defined as number if --execute flag is True')
```
ilaflott, line 23: don't use `sys.exit`, `raise ValueError('text-here')` or smth.

fre/make/create_checkout_script.py
```diff
@@ -18,7 +18,9 @@ def checkout_create(yamlfile, platform, target, no_parallel_checkout, jobs, exec
     run = execute
     jobs = str(jobs)
     pcheck = no_parallel_checkout
-
+    
+    if jobs == 'False' and execute == True:
```
ilaflott, line 22: why is `jobs` a string of a `bool`? Why are we checking if `execute` is `True` using an equality operator? Why can't this be- `if not jobs and execute:` ?

fre/make/tests/null_example/test_checkout_null_yaml.py
ilaflott: this test + the otheres could become a simple:
```
compile_sh_path =  create_checkout_script (<BLAH ARGS>)
assert Path(compile_sh_path).exists()
```

similarly, elsewhere in the code base, if `create_checkout_script` is being used, that returned string object could be actually leveraged by other functions. If it's just `print`, then there is no option to do so

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,48 @@
...
+
+def test_checkout_execute():
+    """
+    check if --execute option works
+    """
+    subprocess.run(["rm","-rf",f"{OUT}/fremake_canopy/test"])
+    create_checkout_script.checkout_create(YAMLFILE,PLATFORM,TARGET,False,2, True,False)
```
ilaflott, line 41: OK so i saw that what brought this PR's checks from red-x to green-check is the change of this functions argument from a boolean value to an integer `2`. This suggests to me that it's hard to track what these arguments are supposed to be when developing the code, so `checkout_create` needs explicit keyword arguments, not positional. 

this line should read like:
```
create_checkout_script.checkout_create(
    yamlfile = YAMLFILE, 
    platform = PLATFORM, 
    target = TARGET, 
    no_parallel_checkout = False, 
    jobs = 2, 
    execute = True, 
    verbose = False )
```

fre/make/tests/null_example/test_checkout_null_yaml.py
ilaflott: +1

fre/make/create_checkout_script.py
singhd789: Then we can keep it nice and just reference `jobs` from the arguments. Also need to find where `jobs=4` is being defined as the default because that's what's supposed to be happening.

fre/make/tests/test_create_checkout.py
```diff
@@ -0,0 +1,27 @@
...
+platform = "gfdl.ncrc5-intel22-classic"
+target = "prod-openmp"
+
+# Set example yaml paths, input directory
+CWD = Path.cwd()
+test_dir = Path("fre/make/tests")
+test_yaml = Path(f"null_example/null_model.yaml")
+
+# Set home for ~/cylc-src location in script
+os.environ["HOME"]=str(Path(f"{CWD}/{test_dir}/configure_yaml_out"))
```
singhd789, line 16: Can we use the `ci.gnu` platform here so `HOME` doesn't need to be reset?

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,48 @@
...
+YAMLFILE = str(Path(f"{TEST_DIR}/null_example/null_model.yaml"))
+
+#set platform and target
+PLATFORM = ["ncrc5.intel23"]
+TARGET = ["debug"]
+
+#set output directory
+# Set home for ~/cylc-src location in script
+#run checkout command
+OUT = f"{TEST_DIR}/configure_yaml_out"
```
singhd789, line 18: This should be changed from `configure_yaml_out` (that was for a my tests for the configure_yaml tool). You could do `checkout_out` or something checkout test related

fre/make/tests/test_create_checkout.py
singhd789, line 16: Change `configure_yaml_out` here to something more for this test, like `checkout_out`

fre/make/tests/test_create_checkout.py
```diff
@@ -0,0 +1,27 @@
...
+from pathlib import Path
+from fre.pp import configure_script_yaml as csy
+
+# Set what would be click options
+experiment = "c96L65_am5f7b12r1_amip"
+platform = "gfdl.ncrc5-intel22-classic"
+target = "prod-openmp"
+
+# Set example yaml paths, input directory
+CWD = Path.cwd()
```
singhd789, line 11: If you use the `ci.gnu` platform with the `TEST_BUILD_DIR` defined for the modelRoot (https://github.com/NOAA-GFDL/fre-cli/blob/a9730e039274739e5dc8710d82dcc09578735d18/fre/make/tests/null_example/platforms.yaml#L20), I don't think you'll even need `CWD`

singhd789 (changes requested):

fre/make/create_checkout_script.py
```diff
@@ -18,7 +18,9 @@ def checkout_create(yamlfile, platform, target, no_parallel_checkout, jobs, exec
     run = execute
     jobs = str(jobs)
     pcheck = no_parallel_checkout
-
+    
+    if jobs == 'False' and execute:
```
singhd789, line 22: I think `jobs` might have a default value here (when you don't explicitly define it, 4 prints out). So it might always be "True" or defined.

kiihne-noaa:
@singhd789 @ilaflott This is now ready for final review

fre/make/tests/null_example/test_checkout_null_yaml.py
```diff
@@ -0,0 +1,83 @@
...
+PLATFORM = ["ci.gnu"]
+TARGET = ["debug"]
+
+#set output directory
+# Set home for ~/cylc-src location in script
+#run checkout command
+OUT = f"{TEST_DIR}/test_run_fremake_multitarget"
+os.environ["TEST_BUILD_DIR"] = OUT
+
+def test_nullyaml_exists():
```
singhd789, line 21: Awesome, thanks for combining the tests into one file! Following other test scripts, just move this one in `fre/make/tests/` and rename it `test_create_checkout.py` (basically replace the other `test_create_checkout.py` with this file)!

singhd789:
Nice changes!! I think besides moving your `test_checkout_null_yaml.py` up one directory (rename `test_create_checkout.py`) and addressing the `if jobs == 'False' and execute:` error check (I think we settled on checking if `jobs` was a boolean and what not instead), this is pretty good.

fre/make/tests/test_create_checkout.py
```diff
@@ -0,0 +1,83 @@
...
+YAMLFILE = str(Path(f"{TEST_DIR}/null_example/null_model.yaml"))
+
+#set platform and target
+PLATFORM = ["ci.gnu"]
+TARGET = ["debug"]
+
+#set output directory
+# Set home for ~/cylc-src location in script
+#run checkout command
+OUT = f"{TEST_DIR}/test_run_fremake_multitarget"
```
singhd789, line 18: Ok sorry, just noticed. I recommend changing `OUT` to `OUT = f"{TEST_DIR}/checkout_out"`, or something checkout related so we know it's associated with this test. It might be checking for output files in the wrong directory here (that path was from the run-fremake test).

fre/make/tests/test_create_checkout.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    Make sure null.yaml is not an empty file
+    """
+    sum(1 for _ in open(f'{YAMLFILE}')) > 1
+    
+def test_checkout_script_exists():
+    """
+    Make sure checkout file exists
+    """
+    subprocess.run(["rm","-rf",f"{OUT}/fremake_canopy/test/null_model_full/src/checkout.sh"])
```
singhd789, line 37: Instead of `subprocess.run`, let's use `shutil.rmtree(OUT)` (for directory removal) of `Path(.......).unlink` (I think it's this) (for file removal) here.

fre/make/tests/test_create_checkout.py
```diff
@@ -0,0 +1,83 @@
...
+                                           verbose = False)
+    
+def test_checkout_no_parallel_checkout():
+    """
+    check if --no_parallel_checkout option works
+    """
+    create_checkout_script.checkout_create(YAMLFILE,
+                                           PLATFORM,
+                                           TARGET,
+                                           no_parallel_checkout = True,
```
singhd789, line 79: I think we'll need to check that actual contents of the file here and add an assert here as well (for file creation). 

If the `no_parallel` option is defined, the checkout script is  going to be different. Maybe we can add a pytest skip here and work on updating this test in the next PR.

singhd789 (approved):

PR #281, opened 2024-11-29
Container base

thomas-robinson:
## Describe your changes
Allows for the containerBase key in the fre make container process.  A container base can be specified.  Has noncorroding updates in [gfdl_msd_schemas](https://github.com/NOAA-GFDL/gfdl_msd_schemas/pull/15) and [mkmf](https://github.com/NOAA-GFDL/mkmf/pull/73)

A test will need to be added when tests are ready

## Checklist before requesting a review

- [x] I ran my code
- [x] I tried to make my code readable
- [x] I tried to comment my code
- [ ] I wrote a new test, if applicable
- [x] I wrote new instructions/documentation, if applicable
- [x] I ran pytest and inspected it's output
- [x] I ran pylint and attempted to implement some of it's feedback

PR #292, opened 2024-12-11
add run-fremake tests

### LINKED ISSUE #226: fre make run-fremake test

rem1776:
## Describe your changes
Adds more tests for the run-fremake command.

For now the container build test is skipped for two reasons:
- podman seems to be running into the same error described here: https://github.com/containers/podman/issues/10802 and may need additional options when the github launches the container (which may or may not be possible)
- the current default base image is 25 gb, and the github runners only have 16gb of total storage (#281 will solve this thankfully)

Fixes #226 

- [x] I ran my code


ilaflott (approved):
This looks fine and dandy to me! Thanks for the clean code @rem1776

Issue #224, opened 2024-10-28
fre make create-compile test


singhd789:
**Is your feature request related to a problem? Please describe.**
fre make create-compile needs an associated test.

**Describe the solution you'd like**
Tests could include 
- [ ] existence of model yaml, platform yaml, and compile yaml
- [ ] successful creation of compile script 
- [ ] successful compilation of null model (if `--execute` used)
- [ ] failures with incorrect compilation info

PR #295, opened 2024-12-16
224.fre make compile test

singhd789:
## Describe your changes
Add create-compile test 

#224 

- [x] I ran my code


kiihne-noaa (commented):
looks good to me! We discussed a number of these changes last week.

fre/make/tests/test_create_compile.py
```diff
@@ -0,0 +1,163 @@
...
+    targ = TARGET[0]
+    yamlfile_path = f"{TEST_DIR}/{NM_EXAMPLE}/{YAMLFILE}"
+
+    # Execute the compile script
+    create_compile_script.compile_create(yamlfile_path, PLATFORM, TARGET, 4, 1, True, False)
+
+    # Check for creation of compile script
+    # Check for FMS directory
+    # Check for log.compile file
+    assert [Path(f"{OUT}/fremake_canopy/test/null_model_full/{plat}-{targ}/exec/compile.sh").exists(),
```
rem1776, line 86: I think this should probably also check for the executable file too, the other files might still be there if it failed.

rem1776 (changes requested):

fre/make/tests/test_create_compile.py
singhd789: Thanks for this catch! You're 100% right

singhd789: So I can't actually run the full compile here because it needs the checked out source code and the makefile (which are not made in this test). Instead of checking for the executable (because it'll fail and won't be there), I can just mark this one as an expected failure but definitely still check for the log file.

singhd789: Ok, I suppose my first though there is already covered by `def test_bad_target_compilelog():`

singhd789: Update: I forgot I can just set an assert equal to False

fre/make/create_compile_script.py
```diff
@@ -79,13 +78,14 @@ def compile_create(yamlfile,platform,target,jobs,parallel,execute,verbose):
                    fremakeBuild.writeBuildComponents(c)
               fremakeBuild.writeScript()
               fremakeBuildList.append(fremakeBuild)
-              click.echo("\nCompile script created at " + bldDir + "/compile.sh" + "\n")
-    if run:
+
+    if execute:
+        click.echo("\nCompile script created at " + bldDir + "/compile.sh" + "\n")
```
ilaflott, line 83: why `click.echo` instead of `print` or `logger` ?

fre/make/tests/test_create_compile.py
```diff
@@ -0,0 +1,165 @@
...
+    plat = PLATFORM[0]
+    targ = TARGET[0]
+    yamlfile_path = f"{TEST_DIR}/{NM_EXAMPLE}/{YAMLFILE}"
+
+    # Execute the compile script
+    create_compile_script.compile_create(yamlfile_path, PLATFORM, TARGET, 4, 1, True, False)
+
+    # Check for creation of compile script, FMS directory,
+    # log.compile file, the executable
+    assert Path(f"{OUT}/fremake_canopy/test/null_model_full/{plat}-{targ}/exec/compile.sh").exists()
```
ilaflott, line 87: `f"{OUT}/fremake_canopy/test/null_model_full/{plat}-{targ}/exec/"` could definitely be assigned to its own variable somewhere. unless you wanna edit the path in several places later upon needed to extend/fix something

fre/make/tests/test_create_compile.py
```diff
@@ -0,0 +1,165 @@
...
+    of compile script creation due to a bad platform passed.
+    """
+    # Set environment variable for use in ci.gnu platform
+    os.environ["TEST_BUILD_DIR"] = OUT
+
+    yamlfile_path = f"{TEST_DIR}/{NM_EXAMPLE}/{YAMLFILE}"
+
+    try:
+        # Create the compile script
+        create_compile_script.compile_create(yamlfile_path, BAD_PLATFORM, TARGET, 4, 1, False, False)
```
ilaflott, line 118: sometimes functions like this:
```
some_func(arg1_val, arg2_val, arg3_val ...)
```

are nicer when written out like this:
```
some_func(arg1 = arg1_val, 
          arg2 = arg2_val, 
          arg3 = arg3_val ...)
```

fre/make/tests/test_create_compile.py
singhd789: the platforms and targets sometimes change (test for bad platforms and targets or multi)

ilaflott (approved):
it works, if needing some cleanup. Granting approval, but the only clarification i'm really wondering: why `click.echo` in `create_compile_script` ?

fre/make/create_compile_script.py
singhd789: Good point. I think it might be print everywhere else. Bennett made it click.echo, which was a cool thing to keep with click functionalities, but I might change it back to print to keep it consistent with the rest of the code

PR #297, opened 2024-12-19
 Update FRE version for 2025 release

singhd789:
## Describe your changes
Just updating the FRE version 

N/A 

- [ ] I ran my code


docs/usage/yaml_framework.rst
```diff
@@ -1,4 +1,4 @@
-In order to utilize FRE 2024.01 tools, a distrubuted YAML structure is required. This framework includes a main model yaml, a compile yaml, a platforms yaml, and post-processing yamls. Throughout the compilation and post-processing steps, combined yamls that will be parsed for information are created. Yamls follow a dictionary-like structure with ``[key]: [value]`` fields. 
+In order to utilize FRE 2025.01 tools, a distrubuted YAML structure is required. This framework includes a main model yaml, a compile yaml, a platforms yaml, and post-processing yamls. Throughout the compilation and post-processing steps, combined yamls that will be parsed for information are created. Yamls follow a dictionary-like structure with ``[key]: [value]`` fields.
```
ceblanton, line 1: I bet we're going to miss updating some of these for 2025.02 and so on.

Maybe rephrase as "In order to utilize FRE tools"

docs/usage/yaml_framework.rst
```diff
@@ -57,10 +57,10 @@ Where each dash indicates a list.
 
 .. code-block:: 
 
-  &version "2024.01"
+  &version "2025.01"
```
ceblanton, line 60: This is only an example of reusable variables, right?

Reusing the fre release number here, and having it be called "version" might give users that it's actually used by FRE do so something. Plus, we'll forget to update it.

fre/tests/test_fre_cli.py
```diff
@@ -24,11 +24,11 @@ def test_cli_fre_option_dne():
 
 def test_fre_version():
     ''' module import flavor of below cli test '''
-    assert '2024.01' == fre.version
+    assert '2025.01' == fre.version
```
ceblanton, line 27: I guess this is ok.. we'll have to update this test every release.

ceblanton (approved):
Thanks! It looks great. I fear we'll forget some of these updates every release, but we can cross that as we go.

fre/tests/test_fre_cli.py
singhd789: Yeah I wasn't sure about this one, definitely wanted to check

docs/usage/yaml_framework.rst
singhd789: Do you mean just doing this instead `&version "(fre release number)"`? Could also do a variation of `&version "xxxx.xx"`?

ceblanton: why have "version" as a user-defined variable example though? Users might think that FRE is using that "version" variable somehow which it is not. Especially if in the example it's 2025.01 users will think it's actually used.

How about another example showing the user-defined variable, "Start_year" maybe.

fre/tests/test_fre_cli.py
ilaflott: updating the version explicitly is the point. if the version is not updated everywhere, the pipeline SHOULD prevent us from publishing.

ilaflott: like- if you are changing either this `assert` statement, or `fre.version`, and you forgot to change it elsewhere, the pipeline should absolutely bonk a dev on the head.

docs/usage/yaml_framework.rst
ilaflott: > Users might think that FRE is using that "version" variable somehow which it is not. Especially if in the example it's 2025.01 users will think it's actually used.

regarding this statement only... https://github.com/search?q=repo%3ANOAA-GFDL%2Ffre-cli+fre.version&type=code

agree that the module version probably shouldn't be used as an instructive lesson in `yaml` configuration, though.

singhd789: > why have "version" as a user-defined variable example though? Users might think that FRE is using that "version" variable somehow which it is not. Especially if in the example it's 2025.01 users will think it's actually used.
> 
> How about another example showing the user-defined variable, "Start_year" maybe.

OH. Yes ok I see, I 100% agree. I can change that tomorrow

ilaflott (approved):

PR #304, opened 2024-12-26
 Fix error message for incorrect platform used

singhd789:
## Describe your changes
If an incorrect platform was passed with the fre make commands, there was a very informative error raised. This fixes the error message to basically say "the platform passed doesn't exists in the platforms.yaml"

- [x] I ran my code

### DIFF
--- a/fre/make/run_fremake_script.py
+++ b/fre/make/run_fremake_script.py
@@ -69,7 +69,7 @@ def fremake_run(yamlfile,platform,target,parallel,jobs,no_parallel_checkout,exec
         if modelYaml.platforms.hasPlatform(platformName):
             pass
         else:
-            raise ValueError (platformName + " does not exist in platforms.yaml")
+            raise ValueError (f"{platformName} does not exist in platforms.yaml")
 
         platform = modelYaml.platforms.getPlatformFromName(platformName)
 
--- a/fre/make/run_fremake_script.py
+++ b/fre/make/run_fremake_script.py
@@ -69,8 +69,7 @@ def fremake_run(yamlfile,platform,target,parallel,jobs,no_parallel_checkout,exec
         if modelYaml.platforms.hasPlatform(platformName):
             pass
         else:
-            raise ValueError(f'{platformName} does not exist in '
-                             f'{modelYaml.combined.get("compile").get("platformYaml")}')
+            raise ValueError (platformName + " does not exist in platforms.yaml")
 
         platform = modelYaml.platforms.getPlatformFromName(platformName)
 
--- a/fre/make/create_compile_script.py
+++ b/fre/make/create_compile_script.py
@@ -56,7 +56,7 @@ def compile_create(yamlfile,platform,target,jobs,parallel,execute,verbose):
          if modelYaml.platforms.hasPlatform(platformName):
               pass
          else:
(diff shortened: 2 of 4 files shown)


fre/make/run_fremake_script.py
```diff
@@ -69,8 +69,7 @@ def fremake_run(yamlfile,platform,target,parallel,jobs,no_parallel_checkout,exec
         if modelYaml.platforms.hasPlatform(platformName):
             pass
         else:
-            raise ValueError(f'{platformName} does not exist in '
-                             f'{modelYaml.combined.get("compile").get("platformYaml")}')
+            raise ValueError (platformName + " does not exist in platforms.yaml")
```
ilaflott, line 72: maybe don't get rid of the f-string? nit.

ilaflott: also, why the code repeat? i assume cutting down on that is a whole PRs worth of work? also nit... for now. 

what makes it tricky for someone who may want to address the code repeat, if it is tricky?

ilaflott (approved):
nothing but nits, but consider them.

fre/make/run_fremake_script.py
singhd789: Ah, I agree about the f-string, I'll put that back in (don't know why I changed that bit). 

Regarding the repeating code - do you mean the fact that the print statement is in each tool?

ilaflott: For dev history (a.k.a. posterity), print statement itself is not really the repeat that i'm noticing, though it does seem duplicated and we need to replace print statements with `logging`. but that's a different thing.

the similarity can be most easily seen across these links:
[`create_checkout_script`](https://github.com/NOAA-GFDL/fre-cli/blob/e699d0807946931e6b5dd213293d4f49d28e2bb2/fre/make/create_checkout_script.py#L60-L99)
[`create_compile_script`](https://github.com/NOAA-GFDL/fre-cli/blob/e699d0807946931e6b5dd213293d4f49d28e2bb2/fre/make/create_compile_script.py#L51-L88)
[`create_docker_script`](https://github.com/NOAA-GFDL/fre-cli/blob/e699d0807946931e6b5dd213293d4f49d28e2bb2/fre/make/create_docker_script.py#L37-L79)
[`create_makefile_script`](https://github.com/NOAA-GFDL/fre-cli/blob/e699d0807946931e6b5dd213293d4f49d28e2bb2/fre/make/create_makefile_script.py#L34-L77)

basically, it looks like we need an abstract base class/method, `script_maker`, and have `makefile_script_creator`, `compile_script_creator`, `checkout_script_creator`, and `docker_script_creator` all inherit from that class or build on the base function.

To a lesser extent, `run_fremake_script` also has a very similar structure, but it's name implies a more distinct purpose, so i'm leaving it out thinking it wouldn't generalize to the other four simply/trivially.

singhd789: ahhhh, ok I understand what you mean now. If you could, can you add to the issue below so we have this suggestion documented there as well? It's great input

https://github.com/orgs/NOAA-GFDL/projects/32/views/1?pane=issue&itemId=81108926&issue=NOAA-GFDL%7Cfre-cli%7C190

PR #307, opened 2024-12-31
add fre dora tools

menzel-r:
## Describe your changes
Adds functionality to interact with `dora` - (we should think about the best way to test this without polluting dora).
These commands will only work with a valid `dora` api token.

- [ ] I ran my code

### DIFF
--- a/fre/analysis/tests/test_subtools.py
+++ b/fre/analysis/tests/test_subtools.py
@@ -1,14 +1,14 @@
 from pathlib import Path
-import pytest
 from subprocess import CalledProcessError
 from tempfile import TemporaryDirectory
 
 from analysis_scripts import available_plugins, UnknownPluginError
 from fre.analysis.subtools import install_analysis_package, list_plugins, run_analysis
+import pytest
 
 
 def make_experiment_yaml(path, name, whitespace="  "):
-    """Creates and experiment yaml configuration file for testing.
+    """Creates an experiment yaml configuration file for testing.
 
     Args:
         path: Path to the experiment yaml file that will be created.
@@ -27,9 +27,9 @@ def test_install_analysis_package():
     url = "github.com/noaa-gfdl/analysis-scripts"
     name = "freanalysis_clouds"
     with TemporaryDirectory() as tmp:
-         install_analysis_package(url, name, tmp)
-         plugins = list_plugins(tmp)
-         assert name in list_plugins(tmp)
+        install_analysis_package(url, name, tmp)
+        plugins = list_plugins(tmp)
+        assert name in list_plugins(tmp)
 
 
 def test_run_analysis():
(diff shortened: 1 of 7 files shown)


fre/analysis/tests/test_subtools.py
```diff
@@ -1,14 +1,14 @@
 from pathlib import Path
-import pytest
 from subprocess import CalledProcessError
 from tempfile import TemporaryDirectory
 
 from analysis_scripts import available_plugins, UnknownPluginError
 from fre.analysis.subtools import install_analysis_package, list_plugins, run_analysis
+import pytest
```
ilaflott, line 7: @menzel-gfdl i'm gonna nit you here more for a good discussion on style regardless of impact. `pylint` doesn't seem to care, but shouldn't the module `import` order be, to steal `C++` syntax for a minute: 
```
<pystdlib> 
<3rd-party external modules>
<sub modules in this package>
```

all sub-sorted alphabetically of course, i can't unsee that now

ilaflott (commented):
looks in the right direction

idk what we can do to avoid polluting `dora`'s database without including some kind of mock DB on `dora`'s end to explicitly address testing cruft.

fre/analysis/tests/test_subtools.py
menzel-gfdl: I always wonder the same thing too.  I guess is the question is, as far as test_subtools.py is concerned, is `fre` a 3rd party module?

menzel-gfdl: https://peps.python.org/pep-0008/#imports

Is what I did compatible with this?

ilaflott: > I always wonder the same thing too. I guess is the question is, as far as `test_subtools.py` is concerned, is `fre` a 3rd party module? I like to view the tests as outside of the package (thus the non-relative imports), but I don't know if I am correct about that.

This is exactly why I thought i'd stir the pot here- I had not thought of, and I agree, that the question to answer is really, "are the tests part of the package?"  In `fre-cli`'s case, for now at least, the tests are considered part of the package. If they're part of the package, then i think the text from the link applies: 

    "Imports should be grouped in the following order: 
     1. Standard library imports. 
     2. Related third party imports. 
     4. Local application/library specific imports."

so by that logic, `analysis_scripts` is third-party in this context... though perhaps this feels a little silly because we work on that package so closely. `pytest` is similarly third-party, but `fre.analysis.subtools` is the local application import. 

just in case it's not clear, this is absolutely a nit and not a blocking issue whatsoever.

menzel-gfdl: Sounds good, thanks!

ilaflott:
i'm not against this in theory, but it's not gonna happen in this specific PR.

goodnight, sweet PR

PR #308, opened 2025-01-02
162.fix yaml overwrites

singhd789:

- [x] I ran my code


ceblanton (commented):
Thank you very much Dana for this needed fix. I can confirm it works properly with my analysis script use case. (It's still in draft so I didn't click approve, plus I didn't really examine the changes)

singhd789:
@ceblanton Thank you for trying it out! At the moment, it combines analysis info, then only combines component info if more than 1 pp yaml is listed. Would other things in a pp yaml need to be combined besides components?

ceblanton:
I see what you mean, this is sort of tricky. 

```
experiments:
  - name: !join [c96L65_, *AM5_VERSION, _amip]
    pp:
      - "yaml_include/pp/amip.c96.yaml"
      - "yaml_include/pp/extras.yaml"
```

With the first pp yaml as:

```
directories:
  <<: *shared_directories
  pp_grid_spec: *GRID_SPEC96

postprocess:
  settings:
    <<: *pp_settings
    pp_start:   *AMIP_START
    pp_stop:    *AMIP_END
    pp_chunk_a: *PP_AMIP_CHUNK96
    pp_components: "atmos atmos_scalar land"
  switches: 
    <<: *pp_switches

  components:
    - type: "atmos_cmip"
      sources: "atmos_month_cmip atmos_8xdaily_cmip atmos_daily_cmip"
      sourceGrid: "cubedsphere"
      xyInterp: *PP_XYINTERP96
      interpMethod: "conserve_order2"
      inputRealm: 'atmos'
```

and the second one as:

```
postprocess:
  components:
    - type: "land_cmip"
      sources: "land_month_cmip"
      sourceGrid: "cubedsphere"
      xyInterp: *PP_XYINTERP96
      interpMethod: "conserve_order1"
      inputRealm: 'land'
```

Would the result be the same if the referenced pp yamls were in the opposite order? like

```
experiments:
  - name: !join [c96L65_, *AM5_VERSION, _amip]
    pp:
      - "yaml_include/pp/extras.yaml"
      - "yaml_include/pp/amip.c96.yaml"
```

singhd789:
I think it would be but let me try it out first!

ceblanton:
I know! It's a real potential problem here.

I think we should be really strict to resolve this potential problem. If the `settings`, `switches` or `directories` or conflicting among the sub-yamls, then I think it should be a hard ERROR.

Actually, this is the same problem that could occur with combining the components. What if the first yaml has a different component definition than the second?

ceblanton:
Meant to say, what if the first pp yaml and the second pp yaml have the same component name (say "atmos") with two different definitions?

ceblanton (approved):
Thank you very much Dana. It's just what we need to make the pp work now.

Issue #317, opened 2025-01-08
fre pp wrapper fails due to 'fre pp trigger' failing because the Cylc scheduler takes 30 seconds to start up


ceblanton:
**Describe the bug**
Tests have shown that the Bronx-23 pp.started FRE 2025 workflows have a repeatable problem.

`fre pp wrapper` runs the `fre pp` tools in order: checkout, configure-yaml, install, run, trigger

When the Cylc scheduler is not yet started, the `run` step starts it. The run step is essentially `cylc play`, which exits 0 before the Cylc scheduler is up and ready to respond to `cylc trigger`.

**To Reproduce**
Any `fre pp wrapper` usage where the Cylc scheduler is not yet running will cause it.

**Expected behavior**
`fre pp run` itself should wait until the Cylc scheduler is actually up before exiting with success.

**Additional context**

PR #318, opened 2025-01-08
Updates to 'fre pp run' to account for delay in starting Cylc scheduler

ceblanton:
## Describe your changes
- check to see if the scheduler is already up; if so exit right away with success
- otherwise, sleep 30 seconds, then confirm the scheduler is actually up
- if not, error 30 seconds was chosen based on hand-testing and might have to be adjusted later

#317

- [x] I ran my code
~~- [ ] I wrote a new test, if applicable~~ (calls to `cylc` are not yet testable presently)

### DIFF
--- a/fre/pp/run_script.py
+++ b/fre/pp/run_script.py
@@ -1,6 +1,7 @@
 ''' fre pp run '''
 
 import subprocess
+import time
 import click
 
 def pp_run_subtool(experiment, platform, target):
@@ -9,9 +10,24 @@ def pp_run_subtool(experiment, platform, target):
     <experiment>__<platform>__<target>
     """
 
+    # Check to see if the workflow is already running
     name = experiment + '__' + platform + '__' + target
+    result = subprocess.run(['cylc', 'scan', '--name', f"^{name}$"], capture_output=True).stdout.decode('utf-8')
+    if len(result):
+        print("Workflow already running!")
+        return
+
+    # If not running, start it
     cmd = f"cylc play {name}"
     subprocess.run(cmd, shell=True, check=True)
+
+    # wait 30 seconds for the scheduler to come up; then confirm
+    print("Workflow started; waiting 30 seconds to confirm")
+    time.sleep(30)
+    result = subprocess.run(['cylc', 'scan', '--name', f"^{name}$"], capture_output=True).stdout.decode('utf-8')
+
+    if not len(result):
+        raise Exception('Cylc scheduler was started without error but is not running after 30 seconds')
 
 @click.command()
 def _pp_run_subtool(experiment, platform, target):


ceblanton:
It's downright shameful that I'm checking only 3 out of 7 of the Ready checklist here...

We don't actually run anything Cylc-related in the CI, though, so how would one test it?

ilaflott (dismissed):
this is less a "request changes" and more a, "request clarification".

The [status script](https://github.com/NOAA-GFDL/fre-cli/blob/main/fre/pp/status_script.py) also has a time-out time. 

Question- instead of trying to ping for the scheduler itself, why not just issue a `status` check until the desired job appears? Why do we need two different time-out times?

ilaflott:
> It's downright shameful that I'm checking only 3 out of 7 of the Ready checklist here...

do a `.md` cross out, surrounding with ~~ to signify "not-applicable". 

> We don't actually run anything Cylc-related in the CI, though, so how would one test it?

given that `wrapper` calls run a workflow, to test it in `fre-cli` sensibly, you're looking at either developing a suitable mock set-up (non-trivial, but containers can/should be used). IMO, that's not the biggest bang-for-buck at this time. We don't have any tests for `fre-workflows` yet, and that repo is about our workflows. We could use `fre-cli` to notch a basic integration test in `fre-workflows`. That'll take some data

fre/pp/run_script.py
```diff
@@ -9,10 +10,25 @@ def pp_run_subtool(experiment, platform, target):
     <experiment>__<platform>__<target>
     """
 
+    # Check to see if the workflow is already running
     name = experiment + '__' + platform + '__' + target
+    result = subprocess.run(['cylc', 'scan', '--name', f"^{name}$"], capture_output=True).stdout.decode('utf-8')
+    if len(result):
+        print("Workflow already running!")
+        return
```
singhd789, line 18: Do you think we should we check/output for beginning task failures that happened in the running workflow? 

For example, I ran an pp.starter sbatch script on gaea. Checking through `cylc scan` on ppan, it shows my experiment is running. However, I had already gotten an email about the pp.starter failing. Soo, if the pp-starter or stage-history task has already failed, the workflow spends quite a bit of time still "running" when really I know the rest of the workflow won't run if those failed already. Instead of waiting 10-15 minutes for a complete failure, maybe we can check those tasks, and either `cylc stop` or print something? 

Hmmm, I guess maybe we don't want a complete stop if only some years fail with the pp-starter task though. Eh, I'm just thinking out loud here, no need to listen to this one. 

ALSO, every once in a while, we might get something like this in doing cylc scan:
```
INFO - Removed contact file for c96L65_am5f7b12r1_amip__gfdl.ncrc5-intel22-classic__prod-openmp (workflow no
    longer running).
WARNING - Workflow not running: c96L65_am5f7b12r1_amip__gfdl.ncrc5-intel22-classic__prod-openmp
```
This was obviously an experiment I wasn't running, but here the stdout in `result` wouldn't be empty (now this doesn't happen often I don't think - and honestly it may have only been because I accidentally did `cylc scan` instead of `cylc scan --name [experiment]`). According the the `cylc scan --help`: 
"These files may left behind if the scheduler did not shut down cleanly)" 

Just wanted to be aware of this...

ceblanton:
> > It's downright shameful that I'm checking only 3 out of 7 of the Ready checklist here...
> 
> do a `.md` cross out, surrounding with ~~ to signify "not-applicable".
> 
> > We don't actually run anything Cylc-related in the CI, though, so how would one test it?
> 
> given that `wrapper` calls run a workflow, to test it in `fre-cli` sensibly, you're looking at either developing a suitable mock set-up (non-trivial, but containers can/should be used). IMO, that's not the biggest bang-for-buck at this time. We don't have any tests for `fre-workflows` yet, and that repo is about our workflows. We could use `fre-cli` to notch a basic integration test in `fre-workflows`. That'll take some data

I think we could have a "dummy" trivial Cylc workflow that could live here in fre-cli, that could do nothing and still test the basic `cylc install`, `cylc run`, and `cylc scan` calls that the `fre pp wrapper` set is exercising.

Sort of like the null model the fre make team is using to test the compile scripts maybe.

ceblanton:
Thanks Ian. Dana, I wish we were already using the external triggers for the history file check, which would present a more natural view of the pp.starting handoff. The way we have it now is just bad.. it doesn't bother me that much because I can imagine how it should be (plus, the proto Shield pp flow.cylc already does this the right way), but it's bad nonetheless and it would drive me crazy too.

fre/pp/run_script.py
singhd789: Yeah, the 20-30 minutes of waiting I'm pretty sure is due to when the workflow is stalled. When that happens, I think doing `cylc scan --state=running --name [name]`, still outputs with the experiment name. That's why I added quotes around running.

ceblanton:
Thanks for letting this one through, you two. Much of what bugs you Dana about this interface is exactly what we need to improve. Silent errors and partial errors (stalled, etc) are so naturally confusing and unpalatable.

We need a unified way for the user to recover when things go wrong, and a way to give users the bad news when necessary. As we get better with Cylc this will look better, but I can't disagree it's not very clear right now.

We have a few bad-news scenarios now. When the pp.starter (the Bronx-submitted job, not the Cylc task) task fails, when when the Cylc workflow stalls, and when it shuts down for good (with removing the workflow share directory).

PR #329, opened 2025-01-23
315.fre list tool

singhd789:

- [x] I ran my code


fre/list_/list_experiments_script.py
```diff
@@ -0,0 +1,71 @@
...
+        print(f"{combined} removed.")
+
+def list_experiments_subtool(yamlfile):
+    """
+    List the post-processing experiments available
+    """
+    # Regsiter tag handler
+    yaml.add_constructor('!join', join_constructor)
+
+    e = "exp_placeholder"
```
ilaflott, line 56: can we make these `"FOO_placeholder"` strings `None` instead?

fre/list_/tests/null_example/compile.yaml
ilaflott, line 1: don't we have a `null.yaml` elsewhere to leverage?

fre/list_/tests/null_example/null_model.yaml
ilaflott, line 1: same comment as other

fre/list_/tests/test_list_experiments_script.py
```diff
@@ -0,0 +1,33 @@
...
+    list_experiments_script.list_experiments_subtool(f"{TEST_DIR}/{NM_EXAMPLE}/{YAMLFILE}")
+
+    #Capture output
+    out,err=capfd.readouterr()
+    if "Post-processing experiments available" in out:
+        assert True
+    else:
+       assert False
+
+@pytest.mark.xfail()
```
ilaflott, line 30: is this test just *expected* to fail right now because of circumstances (e.g. CANNOT succeed in platform)? or is a specific failure *supposed to occur* because of e.g. `raise` within one of the functions?

if former- OK. If latter, consult `pytest` doc on [expected ](https://docs.pytest.org/en/stable/how-to/assert.html#assertions-about-expected-exceptions)

fre/list_/list_experiments_script.py
```diff
@@ -0,0 +1,71 @@
...
+    # Combine model / experiment
+    comb = cy.init_pp_yaml(yml,exp,platform,target)
+    comb.combine_model()
+
+def clean(combined):
+    """
+    Remove intermediate combined yaml.
+    """
+    if Path(combined).exists():
+        os.remove(combined)
```
ilaflott, line 46: this is a nit, but why `os.remove` and not `pathlib.Path.unlink()` ?

fre/list_/tests/null_example/platforms.yaml
```diff
@@ -0,0 +1,20 @@
...
+     RUNenv: [". /spack/share/spack/setup-env.sh", "spack load libyaml", "spack load netcdf-fortran@4.5.4", "spack load hdf5@1.14.0"]
+     modelRoot: /apps
+     container: True
+     containerBuild: "podman"
+     containerRun: "apptainer" 
+     containerBase: "ecpe4s/noaa-intel-prototype:2023.09.25" 
+     mkTemplate: "/apps/mkmf/templates/hpcme-intel21.mk"
+   - name: ci.gnu
+     compiler: gnu
+     mkTemplate: /__w/fre-cli/fre-cli/mkmf/templates/linux-ubuntu-xenial-gnu.mk
```
ilaflott, line 19: this may be changing very soon...

ilaflott (changes requested):
this is a solid effort but i have questions before i click an approve button

fre/list_/list_experiments_script.py
singhd789: They're added to the intermediate combined yaml because the `directories` section of the yaml uses those reusable variables defined. If they're not defined, I thought we might get an "Undefined alias" issue when trying to parse the combined yaml. 

BUT, this comment is good. I now see it'll just define those reusable variables with a literal "None" string. So, we can do this I think!

fre/list_/list_platforms_script.py
```diff
@@ -0,0 +1,75 @@
+"""
+Script combines the model yaml with exp, platform, and target to list experiment information.
+"""
+
+import os
+from pathlib import Path
+import yaml
+import fre.yamltools.combine_yamls as cy
+
+def join_constructor(loader, node):
```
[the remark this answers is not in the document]
singhd789: 100%

fre/list_/list_experiments_script.py
singhd789: Welp, that is because I didn't really know about `unlink`. Good thing to look into, thanks.

fre/list_/list_experiments_script.py
ilaflott: NOT to be hard-to-please... but you mean, `e = None` [the None object](https://docs.python.org/3/library/constants.html#None) 

and not... `e = "None"` (a string), correct?

singhd789: ATM though, the `e` defined in `list_platforms` might need to remain how it is though. I believe the fre make code names the combined yaml file after the model yaml name (so in the intermediate combined yaml, it won't find a "combined-None.yaml" --> it would be looking for "combined-null_model.yaml")

fre/list_/tests/null_example/compile.yaml
singhd789: That we do. I guess I was keeping testing files for different tools together, but you are very right. We can just reference the one already created in `fre/make/tests`

fre/list_/list_platforms_script.py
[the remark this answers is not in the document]
singhd789: In fact, the functions are mostly already defined in the combine_yamls tool. I can reference those functions since I already import functions from that tool.

fre/list_/tests/test_list_experiments_script.py
singhd789: The former. This was just a little test thrown in there for if the model yaml file didn't exist. I do think more tests should go in. I just haven't gotten to it yet.

ceblanton (approved):
Looks great! Thanks Dana.

singhd789:
I just have to add in validation real quick, just btw

ceblanton:
Is this ready?

ceblanton (approved):

singhd789:
> Is this ready?

I added validation for the platforms lister tool but not for the experiment tool yet. But that can be added separately I think

Issue #341, opened 2025-02-04
fre yamltools combine-yamls small bug


ceblanton:
**Describe the bug**
fre yamltools combine-yamls won't combine pp yamls if there are more than one, and no components are set in the first one. Having a pp yaml with no components defined is a little odd and out of the original example.

**To Reproduce**
Try to combine a yaml with more than one pp yaml, with no components set in the first one.

```
experiments:
  - name: !join expname
    pp:
      - pp/settings.yaml
      - pp/c96-clim.yaml
```

```
    comb.merge_multiple_yamls(comb_pp, comb_analysis)
  File "/nbhome/fms/conda/envs/fre-test/lib/python3.9/site-packages/fre/yamltools/combine_yamls.py", line 352, in merge_multiple_yamls
    result[key]["components"] = yf[key]["components"] + result[key]["components"]
KeyError: 'components'
```

**Expected behavior**
Combine even if no components defined in first yaml.

**Additional context**

PR #342, opened 2025-02-04
Combine pp yamls even if the first one does not define any components

ceblanton:
## Describe your changes
Small bug fix

#341 

- [x] I ran my code

### DIFF
--- a/fre/yamltools/combine_yamls.py
+++ b/fre/yamltools/combine_yamls.py
@@ -349,7 +349,10 @@ class init_pp_yaml():
                     if key in yf:
                         if isinstance(result[key],dict) and isinstance(yf[key],dict):
                             if key == "postprocess":
-                                result[key]["components"] = yf[key]["components"] + result[key]["components"]
+                                if 'components' in result['postprocess']:
+                                    result['postprocess']["components"] += yf['postprocess']["components"]
+                                else:
+                                    result['postprocess']["components"] = yf['postprocess']["components"]
         # If only one post-processing yaml listed, do nothing
         # (already combined in 'combine_experiments' function)
         elif pp_list is not None and len(pp_list) == 1:
--- a/fre/yamltools/tests/test_combine_yamls.py
+++ b/fre/yamltools/tests/test_combine_yamls.py
@@ -200,3 +200,117 @@ def test_combined_ppyaml_validation():
     schema = json.loads(s)
 
     validate(instance=yml,schema=schema)
+
+def test_combine_pp_yamls(tmp_path):
+    """
+    Verify yaml combiner functionality by combining
+    a model yaml with 3 pp yamls (2 with components).
+    """
+
+    model = {
+        'experiments' : [
+            {
+                'name' : 'expname',
+                'pp'   : [
+                    'pp1.yaml',
+                    'pp2.yaml',
+                    'pp3.yaml'
+                ]
+            }
+        ]
+    }
+
+    pp1 = {
+        'directories' : {
+            'history_dir': 'one',
+            'pp_dir'     : 'two'
+        },
+        'postprocess' : {
+            'settings' : {
+                'history_segment' : 'three',
+                'pp_start'        : 'four'
+            }
+        }
+    }
+
+    pp2 = {
+        'postprocess' : {
+            'components' : [
+                {
+                    'type'    : 'atmos_cmip',
+                    'sources' : "foo bar" },
+                {
+                    'type'    : 'land',
+                    'sources' : "land_month"}
+            ]
+        }
+    }
+
+    pp3 = {
+        'postprocess' : {
+            'components' : [
+                {
+                    'type'    : 'ocean',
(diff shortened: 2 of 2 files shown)


fre/yamltools/combine_yamls.py
```diff
@@ -349,7 +349,10 @@ def merge_multiple_yamls(self, pp_list, analysis_list):
                     if key in yf:
                         if isinstance(result[key],dict) and isinstance(yf[key],dict):
                             if key == "postprocess":
-                                result[key]["components"] = yf[key]["components"] + result[key]["components"]
+                                if 'components' in result['postprocess']:
+                                    result['postprocess']["components"] += yf['postprocess']["components"]
+                                else:
```
singhd789, line 354: Thanks for this catch. When would `components` not be defined? What else would be specifically defined in a pp yaml?

singhd789, line 354: It might be nice to have a little test for this!

ceblanton: You're so right! I'll do it.

ceblanton: You're right, it's kind of a unusual use. I thought an organization like this possibly makes sense, which has no component information in the first pp yaml

    pp:
      - pp/settings.yaml
      - pp/c96-clim.yaml
      - pp/om5.yaml
      - pp/om5-1x1.yaml
      - pp/om5-sections.yaml

singhd789: Ah ok interesting. So what information would populate the `settings.yaml` (that might not already be in the main model yaml)? Looking at the keys in a pp yaml, we have `fre_properties`, `directories`, and the `postprocess` section. I guess directories is the only one I can see just moving to the pp yaml?

ceblanton: thanks Dana, I don't know if this is proper, but I have this now in the pp/settings.yaml file. I'll agree it's odd that the directories are outside the postprocess part.

```
directories:
  history_dir:      !join [/archive/$USER/, *FRE_STEM, /, *name, /, *platform, -, *target, /, history]
  pp_dir:           !join [/archive/$USER/, *FRE_STEM, /, *name, /, *platform, -, *target, /, pp]
  analysis_dir:     !join [/nbhome/$USER/, *FRE_STEM, /, *name]
  ptmp_dir:         "/xtmp/$USER/ptmp"
  pp_grid_spec:     /work/Niki.Zadeh/mosaic_generation/exchange_grid_toolset/workdir/mosaic_c96om5b04v20240410.20240423.an105/mosaic_c96om5b04v20240410.20240423.an105.tar

postprocess:
  settings:
    history_segment:    P1Y
    pp_start:           '0001'
    pp_stop:            '0010'
    pp_chunk_a:         P5Y
    pp_chunk_b:         P10Y
    pp_components:      atmos_cmip
    site:               ppan

  switches:
    do_statics:                 True
    do_timeavgs:                True
    clean_work:                 True
    do_refineDiag:              False
    do_atmos_plevel_masking:    True
    do_preanalysis:             False
    do_analysis:                False
    do_analysis_only:           False
```

singhd789: Ok I see. So you're just thinking of getting rid of the postprocess section in the model yaml?

ceblanton: Dana, you're so far ahead of me in yaml's that I cannot even fully appreciate what you just wrote there.

I'm not wild about this example, having the directories/pp settings/pp switches in a sub-pp yaml, I just don't know the alternatives.

ceblanton: A day late and a dollar short is my motto.

It was rewarding to write this test and good practice, thanks for the nudge.

ceblanton:
I wondered if comparing (unordered) dictionaries was ill-formed, but it seemed to work as is. We might have to change it later tho.

fre/yamltools/tests/test_combine_yamls.py
```diff
@@ -200,3 +200,117 @@ def test_combined_ppyaml_validation():
...
+    yaml.dump(pp3, file_pp3)
+
+    # combine the yamls
+    # output file is 'combined-expname.yaml' in the current directory
+    cy.consolidate_yamls(tmp_path / 'model.yaml', 'expname', 'platform', 'target', 'pp')
+
+    # open the combined yaml and compare to expected reference
+    file_output = open('combined-expname.yaml', 'r')
+    output = yaml.load(file_output, Loader=yaml.SafeLoader)
+    assert output == combined
```
singhd789, line 313: If I remember correctly `==` doesn't compare the file contents. You might need to use `filecmp` or something like that

singhd789 (approved):

fre/yamltools/tests/test_combine_yamls.py
singhd789: Nevermind - comparing dictionaries!

PR #344, opened 2025-02-05
add conditional skips for build tests and updates to run podman in CI

rem1776:
## Describe your changes
Updates the CI to have the capability to build podman images for one of the unit tests. I had to update the image and give some additional options in the workflow yaml for containers to build properly. It also uses a new base image (hpc-mini) to get around any storage issues. All the podman and baremetal compilation tests will check if the required software is in your path before running, otherwise they will be skipped.

Also fixed these bugs:
- the code for using just the mkmf template name (no path) was missing for containerized builds from my last pr
- the generated createContainer.sh script was using the wrong filenames for previously created .sif/.tar files when it goes to delete them before a build (just needed to add the target as well)

- [x] I ran my code


ilaflott (approved):
looks solid and has a test- approve.

.gitignore
```diff
@@ -185,8 +185,6 @@ fre/tests/test_files/outdir/
 fre/app/generate_time_averages/tests/time_avg_test_files/
 fre/app/regrid_xy/tests/*/
 
-fre/make/tests/*/
```
singhd789, line 188: Was this removed for a particular reason? I think we'd want to ignore folders that are generated by fre make tests here.

rem1776: Yeah it was because i was working on files in `fre/make/tests/compilation/` which matches with this pattern so it was ignoring my changes. Maybe I can change it to like `fre/make/tests/*_out/`? I think that should match up with all the output directories a little closer.

rem1776: should be fixed now

fre/make/gfdlfremake/buildDocker.py
```diff
@@ -38,11 +38,11 @@ def __init__(self,base,exp,libs,RUNenv,target,mkTemplate,stage2base):
         self.stage2base = stage2base
 
         # Set up spack loads in RUN commands in dockerfile
-        if RUNenv == "":
+        if len(RUNenv) == 1 and RUNenv[0] == "":
```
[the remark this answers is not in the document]
singhd789: Other than my own confusion on that bit, this looks good!

.gitignore
singhd789: thanks!

PR #366, opened 2025-02-24
fre make:: add the ability to specify the location of the output container

### LINKED ISSUE #333: Specify container location in fre make
Currently containers are built in the location that `fre run` is called.  There should be a way to specify the location that the container is built.

In the platform, add a containerLocation or similar key and append the container to the end of this location.  

The location should be verified on the host system.  

A defualt value of `./` could be used to install in the current working directory. 

The container should be given rwxr-xr-x privileges. (755 I think)

**Describe alternatives you've considered**

**Additional context**
This is needed so that groups can deploy a single container to a central location for all to use.

uramirez8707:
## Describe your changes
add the ability to specify the location of the output container

Fixes #333 
Requires https://github.com/NOAA-GFDL/gfdl_msd_schemas/pull/21

- [x] I ran my code


fre/make/tests/compilation/test_run_fremake_builds.py
```diff
@@ -81,6 +84,16 @@ def test_run_fremake_container_build():
         no_format_transfer=False, execute=True, verbose=VERBOSE)
     assert Path("null_model_full-debug.sif").exists()
 
+@pytest.mark.skipif(not can_container, reason="missing podman/apptainer")
+def test_run_fremake_container_build():
```
singhd789, line 88: You should name this test something different from the previous test

singhd789 (changes requested):

Issue #371, opened 2025-02-26
fre make container build needs exec script in second stage


thomas-robinson:
**Describe the bug**
The two stage container build in fre make needs to copy over the execution script from the first stage and set up a symbolic link in the second stage.  Here is an example of what the Dockerfile should look like for ESM
```
COPY --from=builder /apps/esm4.5_om5b08_compile_v6_cobv3_mekep/exec /apps/esm4.5_om5b08_compile_v6_cobv3_mekep/exec
RUN mkdir -p /apps/bin \
 && ln -sf /apps/esm4.5_om5b08_compile_v6_cobv3_mekep/exec/execrunscript.sh /apps/bin/execrunscript.sh
```
This can be done by modifying `fre make` to add in the link set up line in the second stage of the Dockerfile.

Additionally, the LD_LIBRARY_PATH of the container is being overwritten in frerun.  frerun needs to be patched to allow for the LD_LIBRARY_PATH of the container to be appended on to the end of the APPTAINER_LD_LIBRARY_PATH.  It's currently set up like this:
`      setenv APPTAINERENV_LD_LIBRARY_PATH "${CRAY_LD_LIBRARY_PATH}:${LD_LIBRARY_PATH}:/opt/cray/pe/lib64:/usr/lib64/libibverbs:/opt/cray/libfabric/1.20.1/lib64:/opt/cray/pals/1.4/lib"`
It needs to have no quotes `"` and have `:\$LD_LIBRARY_PATH` added to the end.  Final should be:
`setenv APPTAINERENV_LD_LIBRARY_PATH ${CRAY_LD_LIBRARY_PATH}:${LD_LIBRARY_PATH}:/opt/cray/pe/lib64:/usr/lib64/libibverbs:/opt/cray/libfabric/1.20.1/lib64:/opt/cray/pals/1.4/lib:\$LD_LIBRARY_PATH`  @rem1776 

**To Reproduce**

**Expected behavior**
The model running

**Additional context**
Tom will do fre make.  Ryan may need to do frerun.

singhd789:
Looks like this is good to close!

PR #372, opened 2025-02-27
Adds a link to the execution script in the container to the second stage of the dockerfile

thomas-robinson:
## Describe your changes
This simply adds a link to execution script in the second stage build of the dockerfile.  This is needed only for compatibility with bronx frerun.  There is no new functionality.
#371 

- [ ] I ran my code


thomas-robinson:
@singhd789 @ceblanton fre 2025.01 will need to be patched so that we can run the ESM45 container with this update.  Bronx will also need to be patched per #371

singhd789 (approved):

PR #373, opened 2025-02-28
climatology Timeavg remake

kiihne-noaa:
## Describe your changes
timeavg fucntion can now accept multiple files. Tests work on two files, but any number should work. Creates a new, combined netcdf file, then runs options with it. Deletes file after output is created. Extra parameters added for fre-python-tools scenario to keep it running as it was before the change. Added versions of original tests that function with multiple files.

- [ Y] I ran my code
- [ Y] I tried to make my code readable
- [ Y] I tried to comment my code
- [ Y] I wrote a new test, if applicable
- [ N] I wrote new instructions/documentation, if applicable
- [ Y] I ran pytest and inspected it's output
- [ Y] I ran pylint and attempted to implement some of it's feedback


mlee03:
@kiihne-noaa, could you change the title to start with "CLIMATOLOGY"?  Thank you!

ilaflott (changes requested):
Can you go through the checklist that other PRs use? The template can be copied from [here](https://raw.githubusercontent.com/NOAA-GFDL/fre-cli/refs/heads/main/.github/pull_request_template.md). It will help you self-review, and then i can do a final-review. 

An example of a thing to consider: do we really need an entire `two_file_test.py`, that looks largely like a carbon copy of `test_generate_time_averages.py`? The similarity suggests that perhaps the new two-file test cases should be a set of tests under `test_generate_time_averages.py`, which would reduce code-repeat, which makes it easier to keep the code up-to-date (fewer files to update). 

It's good work- just needs polish.

Kiihne:
> Can you go through the checklist that other PRs use? The template can be copied from [here](https://raw.githubusercontent.com/NOAA-GFDL/fre-cli/refs/heads/main/.github/pull_request_template.md). It will help you self-review, and then i can do a final-review.
> 
> An example of a thing to consider: do we really need an entire `two_file_test.py`, that looks largely like a carbon copy of `test_generate_time_averages.py`? The similarity suggests that perhaps the new two-file test cases should be a set of tests under `test_generate_time_averages.py`, which would reduce code-repeat, which makes it easier to keep the code up-to-date (fewer files to update).
> 
> It's good work- just needs polish.

Will do on the checklist. For the tests, I had separated them intially for simplicity, and then kept them separate at the end due to file length. trying to balance number of files and not having too long files. Combining them would cause the single test file to be over 450 lines long. Just confirming before I do that?

ceblanton (changes requested):
Let's avoid binary Netcdf files in the repository as they are difficult to remove later, and not trackable by git.

Smallest-possible .cdl files (created by ncgen)

kiihne-noaa:
@ilaflott @ceblanton this has been cleaned up, and all netcdf files removed. Please make more comments or approve in the next couple of days. Thank you.

fre/app/generate_time_averages/generate_time_averages.py
```diff
@@ -9,9 +9,22 @@ def generate_time_average(infile = None, outfile = None,
...
         fre_logger.info(locals()) #input argument details
     exitstatus=1
-
-    #needs a case statement. better yet, smarter way to do this? (TODO)
+        
     myavger=None
+
+    #Use cdo to merge multiple files if present
+    merged = False                            
+    if type(infile).__name__=='list' and len(infile)> 1:   #multiple files case. Generates one combined file
```
ceblanton, line 17: This combining logic seems to be both here and in cdoTimeAverager.py, though I'm not sure this one ever is used.

Are doing the combining here once, or in each of the backend's code (cdo, frenctools, xarray) ?

ceblanton:
Nice! When using the CDO backend, it works well. Thank you! (However, I'm confused about the duplicate sections of combining in generate_time_average.py and cdoTimeAverager.py).

But it doesn't seem to work with the fre-nctools backend at all, which is the back-end FRE has to use due to the calendar. And it exits with success when in fact it isn't working. Can you take a look?

```
> fre app gen-time-averages -i ocean_1x1.000101-000212.tos.nc -i ocean_1x1.000301-000412.tos.nc -o myout.nc --pkg fre-nctools

{'self': frenctoolsTimeAverager( pkg=fre-nctools,                                unwgt=False,                                var=None,                                avg_type=all), 'infile': 'ocean_1x1.000301-000412.tos.nc', 'outfile': 'myout.nc'}
output=b'Usage: timavg [-a] [-b] [-d] [-m] -o ofile file [file ...]\nTime averaging script\n\n    -a          skips "average information does not agree" errors.\n    -b          adds time axis bounds and cell methods (CF convention)\n    -d          turns on command echo (for debugging).\n    -m          average time (instead of end time) for t-axis values\n    -W          suppress warning messages (use with caution).\n    -w wght     minimum fraction of missing data needed for valid data.\n                Defaults to 0.0.\n    -o ofile    name of the output file.  Defaults to "timavg.nc".\n    -z level    If using NetCDF4, use deflation level.  Defaults to\n                input file settings (-1).\n    -s 1|0      If using NetCDF4, use shuffle if 1 and don\'t use if 0.\n                Defaults to input file settings (-1).\n    -h          Display this help and exit.\n    -V          Output version information and exit.\n\n\n    file        list netcdf files, each file will be a time record in\n                the output file (the files should be in chronological\n                order)\n'
success
Finished in total time 0.02 second(s)

>ls myout.nc 
ls: cannot access 'myout.nc': No such file or directory
```

ceblanton (approved):
Avery, thank you for this excellent work. Onward!

ceblanton:
@ilaflott on the "files changed" tab there shows no unresolved conversations but on this page, it says merging blocked because of an Ian requested change. But I can't see it. Arrggg!!

ilaflott (approved):

PR #374, opened 2025-03-01
Various

ceblanton:
## Describe your changes
1. Update schema submodule to constrain analysis configs
2. Some massaging of start/stop times specified as years. (cylc interprets "0001" as a year but not "1")
3. Logging

- [x] I ran my code


ceblanton:
@ian same story on this one. changes we definitely want but not the final update (surprise surprise)

fre/pp/configure_script_yaml.py
```diff
@@ -35,17 +40,29 @@ def validate_yaml(yamlfile):
...
+    logger.info(f"Using yaml schema '{schema_path}'")
     # Load the json schema: .load() (vs .loads()) reads and parses the json in one)
-    with open(schema_path,'r') as s:
-        schema = json.load(s)
+    try:
+        with open(schema_path,'r') as s:
+            schema = json.load(s)
+    except:
+        logger.error(f"Schema '{schema_path}' is not valid. Contact the FRE team.")
+        raise
```
ilaflott, line 50: nit: always fun, figuring out the balance between `logging` and `raise`... [SO food for thought?](https://stackoverflow.com/a/52672315)

fre/pp/configure_script_yaml.py
```diff
@@ -11,12 +11,17 @@
...
 from pathlib import Path
-from jsonschema import validate
+from jsonschema import validate, SchemaError, ValidationError
 import yaml
 import metomi.rose.config
 
 import fre.yamltools.combine_yamls_script as cy
 
+import logging
+logging.basicConfig()
```
ilaflott, line 21: in `fre-cli` the configuration for `logger` gets set in `fre.py`- where it can pick up a `verbose` flag... e.g. with `fre -v yamltools configure-yaml ARGS`, so this doesn't need to be done here, unless you really want the secondary config on top of the base `fre` call

fre/pp/configure_script_yaml.py
```diff
@@ -11,12 +11,17 @@
...
+logger = logging.getLogger(__name__)
```
ilaflott, line 22: nit: everywhere else, we're giving it to `fre_logger`... for flavor!

fre/yamltools/combine_yamls_script.py
```diff
@@ -156,7 +156,7 @@ def consolidate_yamls(yamlfile, experiment, platform, target, use, output):
             yml_dict = get_combined_ppyaml(combined,experiment)
         else:
             yml_dict = get_combined_ppyaml(combined,experiment,output)
-            fre_logger.info(f"Combined yaml file located here: {os.getcwd()}/{output}")
+            fre_logger.info(f"Combined yaml file located here: {output}")
```
ilaflott, line 159: if one strongly desires a full-path as output here, `os.path.resolve()` should work fine

ceblanton: Thanks! In this case`output` is already a full path. The previous print was printing a malformed location.

fre/pp/configure_script_yaml.py
ceblanton: Aha! So there is a central logger, and we're unnecessary re-declaring it in each python file now?

fre/pp/configure_script_yaml.py
ceblanton: thanks! I think I like the tracebacks in general for debugging, and don't like to hide them behind `try / except` blocks.  But then it's unclear what the user should do. So I thought I'd split the baby and report the user message _and_ the full/ugly traceback.

ceblanton:
The tests are failing from some pp yaml validation and usage tests which broke from the schema changes. It's a good catch. I'll update the tests with the schema changes and then they should pass.

fre/pp/configure_script_yaml.py
ilaflott: no you do gotta declare it in each file. that's the suggested basic `logging` usage. the `logging` (how-to)[https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial] isn't so bad a read. there's fancier things we could be doing, i'm sure.

the configuration of `logging` is centralized though, so no need to do `basicConfig` unless you want something tool-specific. 

@Ciheim and I were discussing global v. tool-specific `verbose` the other day, maybe we should have it again!

fre/pp/configure_script_yaml.py
ilaflott: works for me. we'll probably want to encode that as a style-policy for the repo at some point and enforce elsewhere.

ilaflott (approved):
thank you for the yaml updates!

ceblanton:
I was going to try to see if I could get the central logging going in a few minutes, which would be a good idea normally as it sets an example, but thank you for merging. The tests were excellent and easy to adjust. 

You'll teach me the proper central logging way and I'll help fix as I go like you do.

ilaflott:
> I was going to try to see if I could get the central logging going in a few minutes, which would be a good idea normally as it sets an example, but thank you for merging. The tests were excellent and easy to adjust.
> 
> You'll teach me the proper central logging way and I'll help fix as I go like you do.

sorry for merging too quick! I invite you to `rebase` and open another PR for your experiment in central logging though

PR #377, opened 2025-03-04
New subtool: Fre pp histval

Ciheim:

- [x] I ran my code

### DIFF
--- a/fre/pp/frepp.py
+++ b/fre/pp/frepp.py
@@ -189,7 +189,7 @@ def nccheck(file_path, num_steps):
 @pp_cli.command()
 @click.option('--history','-hist', required=True, help="Path to directory containing history files")
 @click.option('--date_string','-d', required=True, help="Date string as written in netCDF (.nc) filename")
-@click.option('--warn', '-w', is_flag=True, default=False, help="Warn mode. Missing diag files and incorrect timesteps will not cause validation to fail. Instead, warnings will be printed in the fre log")
+@click.option('--warn', '-w', is_flag=True, default=False, help="Warn mode. Instead of raising an error, a warning will be printed in the fre log if no diag manifest files are present")
 def histval(history,date_string,warn):
     """ Finds diag manifest files in directory containing history files then runs nccheck to validate timesteps for all files in that directory """
     histval_script.validate(history,date_string,warn)
--- a/fre/pp/histval_script.py
+++ b/fre/pp/histval_script.py
@@ -56,7 +56,9 @@ def validate(history,date_string,warn):
                            f"{history}",
                            f"{date_string}.{filename}.tile{tile_num}.nc")
             else:
-                filepath = os.path.join(history,date_string+'.'+filename+'.nc')
+                filepath = os.path.join(
+                           f"{history}",
+                           f"{date_string}.{filename}.nc")
 
             result = ncc.check(filepath,info[filename][0])
 


fre/pp/ncvalidate_script.py
```diff
@@ -0,0 +1,60 @@
...
+import re
+import logging
+from fre.pp import nccheck_script as ncc
+
+fre_logger = logging.getLogger(__name__)
+
+# Mega manifest sounds cool... it'll just be all of the data from the diag_manifests combined in list form
+mega_manifest=[]
+mismatches=[]
+levels={}
```
ceblanton, line 18: Must these 3 variables be global, or could they be defined within `validate`?

fre/pp/frepp.py
```diff
@@ -184,6 +185,14 @@ def nccheck(file_path, num_steps):
     """ - Check that a netCDF (.nc) file contains expected number of timesteps - """
     nccheck_script.check(file_path,num_steps)
 
+#fre pp histval
+@pp_cli.command()
+@click.option('--history','-hist', required = True, help = "Path to directory containing history files")
+@click.option('--date_string','-d', required = False, help = "Date string as written in netCDF (.nc) filename")
```
ceblanton, line 191: we need this to be required for a sane world...

fre/pp/histval_script.py
```diff
@@ -0,0 +1,62 @@
...
+    # Find diag_manifest files and add to mega_manifest
+    files = os.listdir(history)
+    for _file in files:
+        if _file[-1].isdigit() and 'diag_manifest' in _file and not _file.startswith('.'):
+            filepath = os.path.join(history,_file)
+            with open(filepath, 'r') as f:
+                fre_logger.info(f" Grabbing data from {filepath}")
+                data = yaml.safe_load(f)
+                mega_manifest.append(data)
+
```
ceblanton, line 32: If there are no manifests found, consider it an error.

fre/pp/histval_script.py
```diff
@@ -0,0 +1,62 @@
...
+            try:
+                filepath = glob.glob(os.path.join(history,'*'+date_string+'*'+filename+'*.nc'))[0]
+            except:
+                sys.exit("Check date string for correctness")
+        else:    
+            filepath = glob.glob(os.path.join(history,'*'+filename+'*.nc'))[0]
+        ncc.check(filepath,levels[filename])
+        result = ncc.check(filepath,levels[filename])
+        if result==1:
+            fre_logger.info(f" Timesteps found in {filepath} differ from expectation in diag manifest")
```
ceblanton, line 52: This should be reported as an `ERROR` not `INFO`

fre/pp/ncvalidate_script.py
```diff
@@ -0,0 +1,60 @@
...
+            filepath = glob.glob(os.path.join(history,'*'+filename+'*.nc'))[0]
+        ncc.check(filepath,levels[filename])
+        result = ncc.check(filepath,levels[filename])
+        if result==1:
+            fre_logger.info(f" Timesteps found in {filepath} differ from expectation in diag manifest")
+            mismatches.append(_file)
+
+    #Error Handling
+    if mismatches:
+        sys.exit(str(len(mismatches))+ " files contain an unexpected number of timesteps." + "\n".join(mismatches))
```
ceblanton, line 57: I don't think we should be using `sys.exit` here but I can't explain why. I think raising an exception is better. @ilaflott could give a better answer.

Also, please add a newline for the user message. e.g. you're missing one here:

```
11 files contain an unexpected number of timesteps.land_month_inst
land_dust
river_month_inst
iceberg_month
atmos_co2_month
atmos_global_cmip
land_ddep_diurnal
land_ddep
ocean_Bering_Strait
ocean_Drake_Passage
ocean_Indonesian_Throughflo

File: fre/pp/ncvalidate_script.py [hunk 6]
```diff
@@ -0,0 +1,60 @@
...
+                filepath = glob.glob(os.path.join(history,'*'+date_string+'*'+filename+'*.nc'))[0]
+            except:
+                sys.exit("Check date string for correctness")
+        else:    
+            filepath = glob.glob(os.path.join(history,'*'+filename+'*.nc'))[0]
+        ncc.check(filepath,levels[filename])
+        result = ncc.check(filepath,levels[filename])
+        if result==1:
+            fre_logger.info(f" Timesteps found in {filepath} differ from expectation in diag manifest")
+            mismatches.append(_file)
```
ceblanton, line 53: you want `filename` here not `_file`

ceblanton (changes requested):

fre/pp/histval_script.py
```diff
@@ -0,0 +1,62 @@
...
+            expected_timelevels = mega_manifest[y]['diag_files'][x]['number_of_timelevels']
+            levels.update({str(filename):expected_timelevels})
+
+    # Run nccheck to compare actual timelevels to expected levels found in mega manifest
+    for filename in levels:
+        if date_string:
+            try:
+                filepath = glob.glob(os.path.join(history,'*'+date_string+'*'+filename+'*.nc'))[0]
+            except:
+                sys.exit("Check date string for correctness")
```
ilaflott, line 46: instead of `sys.exit`, please use `raise`

fre/pp/ncvalidate_script.py
ilaflott: this is correct RE `sys.exit` v. `raise`, `raise` lets one bring up an `Exception` of a particular type, consult the [built-ins](https://docs.python.org/3/library/exceptions.html). `sys.exit` ALWAYS raises [`SystemError`](https://docs.python.org/3/library/exceptions.html#SystemError), and requires an explicit `import`

fre/pp/ncvalidate_script.py
```diff
@@ -0,0 +1,60 @@
...
+from fre.pp import nccheck_script as ncc
+
+fre_logger = logging.getLogger(__name__)
+
+# Mega manifest sounds cool... it'll just be all of the data from the diag_manifests combined in list form
+mega_manifest=[]
+mismatches=[]
+levels={}
+
+def validate(history,date_string):
```
ilaflott, line 20: why is this so similar to `histval_script`... ? can this be re-factored into a common functionality?

fre/pp/ncvalidate_script.py
```diff
@@ -0,0 +1,60 @@
...
+    """ Compares the number of timesteps in each netCDF (.nc) file to the number of expected timesteps as found in the diag_manifest file(s) """
+
+    # Find diag_manifest files and add to mega_manifest
+    files = os.listdir(history)
+    for _file in files:
+        if _file[-1].isdigit() and 'diag_manifest' in _file and not _file.startswith('.'):
```
ilaflott, line 26: `if CONDITION_1 and CONDITION_2 and ... CONDITION_N:` style things often read better like:
```
if any( [CONDITION_1,
         CONDITION_2,
         ...
         CONDITION_N] ):
```

fre/pp/ncvalidate_script.py
```diff
@@ -0,0 +1,60 @@
...
+        for x in range(len(mega_manifest[y]['diag_files'])):
+            filename = mega_manifest[y]['diag_files'][x]['file_name']
+            expected_timelevels = mega_manifest[y]['diag_files'][x]['number_of_timelevels']
+            levels.update({str(filename):expected_timelevels})
+
+    # Run nccheck to compare actual timelevels to expected levels found in mega manifest
+    for filename in levels:
+        if date_string:
+            try:
+                filepath = glob.glob(os.path.join(history,'*'+date_string+'*'+filename+'*.nc'))[0]
```
ilaflott, line 44: consider `f-string`?  and maybe breaking up heavily nested calls over a few lines? 

```
glob.glob( 
            os.path.join( 
                         f'{history}*{date_string}*{filename}*.nc' ) ) [0]
```

Ciheim: Done!

fre/pp/ncvalidate_script.py
Ciheim: Fixed this

fre/pp/histval_script.py
Ciheim: Done!

fre/pp/ncvalidate_script.py
Ciheim: Fixed!

fre/pp/ncvalidate_script.py
Ciheim: Made the change

fre/pp/histval_script.py
Ciheim: Done!

fre/pp/histval_script.py
Ciheim: Done!

ceblanton (changes requested):
Some kind of test needed. Minimal test, with tiny cdf Netcdf example.

fre/pp/histval_script.py
```diff
@@ -66,12 +70,19 @@ def validate(history,date_string):
                 mismatches.append(filepath)
 
     #Error Handling
-    if len(mismatches)==1:
-        raise ValueError("\n" + str(len(mismatches))+ " file contains an unexpected number of timesteps:\n" + "\n".join(mismatches))
-    elif len(mismatches) > 1:
-        raise ValueError("\n" + str(len(mismatches))+ " files contain an unexpected number of timesteps:\n" + "\n".join(mismatches))
-    else:
-        return(0)
+    if not warn:
```
ceblanton, line 73: This should always be an error. FRE has to use the `warn` mode because users might not have diag manifests yet, but if the user does have diag manifests, a failure from `fre pp histval` should be an actual error, so that Cylc can stop and insist the user resolves it

ceblanton (changes requested):
Almost there...

ceblanton (changes requested):
It looks good, but the tests are failing. Can you fix?

fre/pp/histval_script.py
Ciheim: Thanks, fixed

Ciheim:
*** Last commit does not address error handling. Sorry. Forgot that was my last commit. ***

Ciheim:
> *** Last commit does not address error handling. Sorry. Forgot that was my last commit. ***

eh nvm.. I guess it did technically address the way an error is handled...

ceblanton (approved):
Excellent work, Ciheim.

fre/pp/histval_script.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # Mega manifest sounds cool... it'll just be all of the data from the diag_manifests combined in list form
+    mega_manifest=[]
+    mismatches=[]
+    info={}
+
+    # Find diag_manifest files and add to mega_manifest
+    files = os.listdir(history)
+    diag_count = 0
+    for _file in files:
```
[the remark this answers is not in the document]
Ciheim: Whoops.. still learning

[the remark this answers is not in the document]
ceblanton: me too. why not just use `file` I say? It seems to work

PR #383, opened 2025-03-19
Remove standard deviation feature from fre pp gen-time-averages

ceblanton:
Timeaveraging itself is complicated enough!
Standard deviation calculation could be its own tool in the future.

- [x] I ran my code

### DIFF
--- a/fre/app/freapp.py
+++ b/fre/app/freapp.py
@@ -103,17 +103,13 @@ def mask_atmos_plevel(infile, outfile, psfile):
               help = "Type of time average to generate. \n \
                      currently, fre-nctools and fre-python-tools pkg options\n \
                      do not support seasonal and monthly averaging.\n")
-@click.option("-s", "--stddev_type",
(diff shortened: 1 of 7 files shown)


ceblanton:
@kiihne-noaa I was starting to look at your updates and became confused about whether we were doing timeaverages or standard deviations. Given how challenging climatologies are, I thought we should reduce the complexity of the tool for now, by removing the standard deviation calculations. (@ilaflott agreed in principle.)

kiihne-noaa:
> @kiihne-noaa I was starting to look at your updates and became confused about whether we were doing timeaverages or standard deviations. Given how challenging climatologies are, I thought we should reduce the complexity of the tool for now, by removing the standard deviation calculations. (@ilaflott agreed in principle.)

Looks good to me. I do not think it should mess with my changes very much, though I will need to make a number of minor adjustments. once you have merged to main, I will merge into my branch and make those adjustments.

kiihne-noaa (approved):

PR #384, opened 2025-03-20
Require python 3.10 and remove cylc-uiserver as a dependency

ceblanton:
## Describe your changes
Python 3.9 is too old (not receiving security updates) and probably causes fre analysis install issues.

Hand testing showed cylc-uiserver was causing the solver to always use 3.9.

- [ ] I ran my code

### DIFF
--- a/.github/workflows/create_test_conda_env.yml
+++ b/.github/workflows/create_test_conda_env.yml
@@ -45,7 +45,7 @@ jobs:
           export PATH="$path_save"
 
           # install genbadge to make badge from coverage/test stats
-          pip install genbadge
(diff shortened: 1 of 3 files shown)


ceblanton:
Woohoo! Ian I can't believe you practically eyeballed both problems successfully here. Pinning python to 3.10 fixed the sql build problem, and pip installing defusedxml resolved the genbadge problem.

ilaflott (approved):
So happy we just needed `defusedxml` in addition to `genbadge`, thanks for giving that approach a shot. 

Also, thrilled to be moving beyond `3.9`. Cheers!

PR #387, opened 2025-03-26
 386. Add `fre list pp-components` tool

singhd789:
## Describe your changes
`fre list` tool that lists the post-processing component that has `postprocess_on: True`

- [x] I ran my code


fre/list_/list_pp_components_script.py
```diff
@@ -0,0 +1,52 @@
...
+    yml_dict = quick_combine(yamlfile,e,p,t,"pp",output=None)
+
+    # set logger level to INFO
+    former_log_level = fre_logger.level
+    fre_logger.setLevel(logging.INFO)
+
+    # log the experiment names, which should show up on screen for sure
+    fre_logger.info("Components to be post-processed:")
+    for i in yml_dict["postprocess"]["components"]:
+        if i.get("postprocess_on") is True:
```
ilaflott, line 47: `FOO is True` is generally discouraged for boolean condition checks. If `i` is a dictionary, then `if i.get("FOO") is not None:` should be used instead if what we need to check is the existence of a non-empty `return`, and just `i.get("FOO")` if the expected data type retrieved is a `bool`

ceblanton (approved):
very nice, thanks again Dana. fre list all the way

PR #397, opened 2025-04-16
379.fix list tools

singhd789:

- [ ] I ran my code

### DIFF
--- a/fre/list_/list_experiments_script.py
+++ b/fre/list_/list_experiments_script.py
@@ -1,75 +1,50 @@
 """
 Script combines the model yaml with exp, platform, and target to list experiment information.
 """
-
-import logging
-fre_logger = logging.getLogger(__name__)
-
 from pathlib import Path
+import logging
+from fre.yamltools import pp_info_parser as ppip
 
-# this brings in the yaml module with the join_constructor
-# this is defined in the __init__
-from fre.yamltools import *
-
-
-from fre.yamltools.helpers import yaml_load
-from fre.yamltools import combine_yamls as cy
-
-# To look into: ignore undefined alias error msg for listing?
-# Found this somewhere but don't fully understand yet
-#class NoAliasDumper(yaml.SafeDumper):
-#    def ignore_aliases(self, data):
-#        return True
+fre_logger = logging.getLogger(__name__)
 
 def quick_combine(yml, exp, platform, target):
     """
     Create intermediate combined model and exp. yaml
     This is done to avoid an "undefined alias" error
     """
-    # Combine model / experiment
-    # note, this needs combine_yamls.py, instead of it's successor for now
-    comb = cy.init_pp_yaml(yml,exp,platform,target)
-    comb.combine_model()
 
-def remove(combined):
-    """
-    Remove intermediate combined yaml.
-    """
-    if Path(combined).exists():
-        Path(combined).unlink()
-        fre_logger.info(f"Intermediate combined yaml {combined} removed.")
-    else:
-        raise ValueError(f"{combined} could not be found to remove.")
+    # Combine model / experiment
+    yamldict = ppip.InitPPYaml(yml,exp,platform,target)
+    model_yml_dict = yamldict.combine_model()
+
+    return model_yml_dict
 
 def list_experiments_subtool(yamlfile):
     """
     List the post-processing experiments available
     """
+    # set logger level to INFO
+    former_log_level = fre_logger.level
+    fre_logger.setLevel(logging.INFO)
 
     e = "None"
     p = "None"
     t = "None"
 
-    combined = f"combined-{e}.yaml"
-
     # Combine model / experiment
-    quick_combine(yamlfile,e,p,t)
-    
-    # load the yaml we made
-    c = yaml_load(combined)
+    (yaml_info, yml_dict) = quick_combine(yamlfile, e, p, t)
 
-    # set logger level to INFO
-    former_log_level = fre_logger.level
-    fre_logger.setLevel(logging.INFO)
+#    # Validate combined yaml information
+#    frelist_dir = Path(__file__).resolve().parents[2]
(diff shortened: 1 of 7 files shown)


singhd789:
Note: fre list platforms and fre list pp-components combine yaml information and validates

HOWEVER, there will be a little bit of an issue when trying to validate fre list exps yaml information. This is because not all information is combined with this tool. It essentially just read the model yaml to list the experiments available to be post-processed. It is used like fre list exps -y [model yaml file].

The only validation schema are currently fre_make.json and fre_pp.json, which include more information and set some keys as required (keys we wouldn't need for just reading the model yaml). This might mean we need to add a model_yaml.json schema.

fre/list_/list_experiments_script.py
```diff
@@ -1,75 +1,41 @@
...
-    quick_combine(yamlfile,e,p,t)
-    
-    # load the yaml we made
-    c = yaml_load(combined)
-
     # set logger level to INFO
     former_log_level = fre_logger.level
     fre_logger.setLevel(logging.INFO)
 
+    exp = "None"
```
ilaflott, line 20: can we use `None` instead of a string instead?

fre/list_/list_experiments_script.py
```diff
@@ -1,75 +1,41 @@
...
-    combined = f"combined-{e}.yaml"
-
-    # Combine model / experiment
-    quick_combine(yamlfile,e,p,t)
-    
-    # load the yaml we made
-    c = yaml_load(combined)
-
     # set logger level to INFO
     former_log_level = fre_logger.level
```
ilaflott, line 17: let's move the temp logging level setting to right before the `logger.info` statements below. I think we should let the user decide the verbosity of the yaml-combining steps, and force `INFO` only on the list part.

fre/list_/list_platforms_script.py
```diff
@@ -2,97 +2,42 @@
...
+    yml_dict = cy.consolidate_yamls(yamlfile = yamlfile,
+                                    experiment = exp,
+                                    platform = platform,
+                                    target = target,
+                                    use = "compile",
+                                    output = None)
 
     # Validate the yaml
-    validate_yaml(yml)
+    frelist_dir = Path(__file__).resolve().parents[2]
```
ilaflott, line 33: let's call this `fre_pkg_dir`, and so this should probably be `.parents[1]`, so that `/fre` is left in the path.

fre/list_/tests/test_list_experiments_script.py
```diff
@@ -34,33 +36,14 @@ def test_exp_list(caplog):
...
-    ''' test intermediate combined yaml was cleaned at end of listing '''
-    assert not Path(f"./combined-{EXP_NAME}.yaml").exists()
-
-# Test individual functions operating correctly: combine and clean
-def test_correct_combine():
-    ''' test that combined yaml includes necessary keys '''
-    p = "None"
-    t = "None"
+# Test validation
+@pytest.mark.skip(reason='cannot validate with current schema at the moment')
```
ilaflott, line 40: why can't the current schema validate? a little extra detail here would go a long way

fre/list_/tests/test_list_platforms_script.py
```diff
@@ -45,50 +48,21 @@ def test_platforms_list(caplog):
...
     # Combine model / experiment
-    list_platforms_script.quick_combine(yamlfile_path,PLATFORM,TARGET)
-    assert Path(f"{TEST_DIR}/{NM_EXAMPLE}/combined-{EXP_NAME}.yaml").exists()
+    list_platforms_script.list_platforms_subtool(f"{TEST_DIR}/{NM_EXAMPLE}/{YAMLFILE}")
 
-    comb_yamlfile = f"{TEST_DIR}/{NM_EXAMPLE}/combined-{EXP_NAME}.yaml"
-    with open(comb_yamlfile, 'r') as yf:
-        y = yaml.load(yf,Loader=yaml.Loader)
+    validate = ["Validating YAML information...",
+                "     YAML dictionary VALID."]
```
ilaflott, line 60: lets get some actual `platforms` output to compare with.

ilaflott (commented):
early prelim feedback! to keep things moving

fre/list_/tests/test_list_platforms_script.py
singhd789: of even further, a separate test file for `fre/yamltools/helpers.py`

ilaflott (changes requested):
definite improvements, definitely getting there

fre/list_/tests/test_list_platforms_script.py
```diff
@@ -45,50 +49,55 @@ def test_platforms_list(caplog):
...
-    assert list_platforms_script.validate_yaml(y)
-    #assert "Intermediate combined yaml VALID" in caplog.text
+    for i in validate:
+        assert i in caplog.text
 
+    for record in caplog.records:
+        record.levelname == "INFO"
 
-#def test_not_valid_yaml():
+def test_yamlcontent_valid():
```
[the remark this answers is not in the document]
singhd789: I think I'll open another issue to move this :+1:

ilaflott:
> The Codecov report says "All modified and coverable lines are covered by test", so I'm not 100% sure why the score went down / why it's failing yet

Codecov is using a logical "and" when it describes "lines", ergo, every line you modified in this PR is covered by a test. That's the first default code-coverage req by this service, and the `codecov/patch` green-check shows us it passed.

But, there's a second default code-coverage req- overall coverage of the codebase must go up, not down! 
because a near-fully-covered file was deleted (`fre/yamltools/combine_yamls`), overall project coverage went down! and so `codecov/project` shows a red-x.

I'm uninterested in punishing contributors for deleting old-and-unused code, even if the overall proj coverage goes down. metrics are always flawed, and this is showing us how code-coverage isn't going to ever tell a perfect story without more context. so we should figure out how to config the codecov checks to do what we want... issue coming shortly!

ilaflott (approved):

Issue #280, opened 2024-11-29
Remove the need for -npc when using a container platform


thomas-robinson:
**Is your feature request related to a problem? Please describe.**
Container platforms require the `-npc` for no parallel checkout when creating the checkout script.  There is no check for this and fre does not just do this for you.  If you don't add `-npc` the container build will fail in a mysterious way

**Describe the solution you'd like**
I would like the default to be no parallel checkout, at least for the container platforms.

**Describe alternatives you've considered**
We could also provide a check with an error message if `-npc` is not included, but it seems like podman container building will not support the parallel checkout the way we have them written.

**Additional context**

singhd789:
This should be easy to implement. Should we even keep the non-parallel checkout option for non-container use? Seems like no one would use it, haha.

PR #398, opened 2025-04-17
Removes parallel checkout for containers

thomas-robinson:
## Describe your changes
Sets pc to "" in the creation of the checkout script for a container to make sure that there is no parallel checkout
#280 

- [x] I ran my code


fre/make/create_checkout_script.py
```diff
@@ -110,6 +110,7 @@ def checkout_create(yamlfile, platform, target, no_parallel_checkout, jobs, exec
             src_dir = platform["modelRoot"] + "/" + fremake_yaml["experiment"] + "/src"
             bld_dir = platform["modelRoot"] + "/" + fremake_yaml["experiment"] + "/exec"
             tmp_dir = "tmp/"+platform_name
+            pc = "" #Set this way because containers do not support the parallel checkout
```
singhd789, line 113: You'll need to set the same thing in the `run-fremake` tool as well

singhd789:
Thanks for this Tom!

fre/make/create_checkout_script.py
singhd789, line 113: You'll need to set the same thing in the `run-fremake` tool as well

singhd789 (changes requested):

thomas-robinson:
Really the full tool should just be calling the make checkout script code.  That's for future consideration.

singhd789:
> Really the full tool should just be calling the make checkout script code. That's for future consideration.

Yes it should!! It would be a lot easier to edit the scripts if we didn't have to change it in both places.

That could be the next edit as I don't think that would take long to do at all!

singhd789:
@thomas-robinson Sorry, one more thing - if we make the non-parallel checkout the default for container things, we should also remove it from the documentation / guide (https://github.com/NOAA-GFDL/fre-cli/blob/d376d65b2b288202959a37a22645e91bcb552968/docs/usage/build_fms_model.rst#L50)

docs/usage/build_fms_model.rst
singhd789, line 1: Just a few more instances of `-npc` to remove in here but besides that, looks good to me!

singhd789 (approved):
Changes look good, thanks! Slight fix Ian made for failing tests (click version update caused issues). Merge that in and we're good to go!

PR #402, opened 2025-04-18
394.split.netcdf

cwhitlock-NOAA:
Links to https://github.com/NOAA-GFDL/fre-workflows/issues/51

Moving split_netcdf to fre-cli for better code grouping and testing; new function for doing the variable splitting with xarray.

Completed:
 - Decently comprehensive list of non-diagnostic vars + detection within the file

Known bugs: 
- time_bnds units are not being detected and written out by xarray

Reference for the units issue as it applies to time/time_bnds: https://github.com/pydata/xarray/issues/8368


cwhitlock-NOAA:
Okay - after talking with Chris, I can clarify what we are going for with the two-tool structure: 

We would like to have one tool that follows the unix philosophy (https://en.wikipedia.org/wiki/Unix_philosophy) and splits a single netcdf file with a minimum of fuss like split_ncvars.pl in fre_nctools, and one tool that does the fre-workflows specific stuff that we need (mostly checks on file existence, path construction and yaml parsing). BOTH of these tools should be exposed to users via the click bindings. 

This roughly maps onto what split_netcdf (the fre-workflows specific tool) and split_vars_xarray (the unixlike tool) are already doing, but they need a bit of refactoring to suit that particular vision.

cwhitlock-NOAA:
[ ] flow.cylc passes in both component and source:history file information and CLEARLY LABELS EACH
[ ] update yaml parsing to reflect changes in how the history files are specified
[ ] add schema update for the variable list specification
[ ] check for other places in flow.cylc where "component" is used to mean "history file"

PR #421, opened 2025-05-06
Updates for proposed changes to PP yaml format

J-Lentz:
## Describe your changes
The code, documentation, tests, and templates have been updated in accordance with the changes to the PP yaml format proposed in NOAA-GFDL/gfdl_msd_schemas#28:
- Replace `pp_chunk_a` and `pp_chunk_b` with an array called `pp_chunks`
- Replace `cumulative` and `script_frequency` with `script_type` and `chunk_size` in analysis script configurations.

This should be tested in conjunction with NOAA-GFDL/gfdl_msd_schemas#28 and NOAA-GFDL/fre-workflows#69.

## Issue ticket number and link
#409

## Checklist before requesting a review

- [ ] I ran my code
- [x] I tried to make my code readable
- [x] I tried to comment my code
- [x] I wrote a new test, if applicable
- [x] I wrote new instructions/documentation, if applicable
- [ ] I ran pytest and inspected it's output
- [ ] I ran pylint and attempted to implement some of it's feedback
- [x] No print statements; all user-facing info uses logging module

PR #425, opened 2025-05-07
Do not quote lists in rose-suite.conf

ceblanton:
## Describe your changes
#421 caused 'fre pp configure-yaml' to write incorrect lists to rose-suite.conf. The double quoting is not allowed.

```
PP_CHUNKS='['P1Y']'
```

This update adds a check in `quote_rose_values` for whether the config item is a `list` and if so skips the quoting.

- [x] I ran my code

### DIFF
--- a/fre/pp/configure_script_yaml.py
+++ b/fre/pp/configure_script_yaml.py
@@ -73,8 +73,6 @@ def rose_init(experiment,platform,target):
     rose_suite = metomi.rose.config.ConfigNode()
     # disagreeable; these should be optional
     rose_suite.set(keys=['template variables', 'DO_ANALYSIS_ONLY'],  value='False')
-    rose_suite.set(keys=['template variables', 'DO_MDTF'],  value='False')
-    rose_suite.set(keys=['template variables', 'PP_DEFAULT_XYINTERP'],  value='0,0')
 
     # set some rose suite vars
     rose_suite.set(keys=['template variables', 'EXPERIMENT'], value=f'"{experiment}"')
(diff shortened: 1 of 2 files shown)


fre/pp/configure_script_yaml.py
```diff
@@ -73,8 +73,6 @@ def rose_init(experiment,platform,target):
     rose_suite = metomi.rose.config.ConfigNode()
     # disagreeable; these should be optional
     rose_suite.set(keys=['template variables', 'DO_ANALYSIS_ONLY'],  value='False')
-    rose_suite.set(keys=['template variables', 'DO_MDTF'],  value='False')
-    rose_suite.set(keys=['template variables', 'PP_DEFAULT_XYINTERP'],  value='0,0')
```
singhd789, line 77: Were these not needed anymore?

singhd789:
Besides a clarification, great catch, looks good!

fre/pp/configure_script_yaml.py
ceblanton: Yeah, good question. I confess I wasn't sure, but I haven't seen a cylc validation error yet so I think it's safe. I don't think we need them any more (and I personally never liked either one of these options!)

ceblanton: Huh, on second check, you're absolutely right. Both of those variables are mentioned in fre-workflows.

ilaflott (changes requested):
unit test.

fre/pp/configure_script_yaml.py
ceblanton: Whew, thanks for the excellent save here, Dana. This would have been a problem. I added those 2 lines back, but still dislike them lol.

We'll remove them soon in a more orderly way.

singhd789:
Ok good point Ian - sometimes when it's so simple I forget about unit tests 😬

ceblanton:
Thanks both of you for catches on this "simple" pr.

ilaflott (approved):
❤️❤️❤️   i love it   ❤️❤️❤️

PR #428, opened 2025-05-09
Another attempt for adjusting fre app regrid

ceblanton:
## Describe your changes
1. `fre app regrid` is still reading the old Rose app configuration for now. This update passes the configuration in as an option for now. We'll refactor this to use the yamls very soon.
2. A list of unregriddable variables is stored and applied.
3. Remove undesired default xyinterp option

Needed for https://github.com/NOAA-GFDL/fre-workflows/pull/75

- [x] I ran my code


ilaflott:
OK. Brain hurts. But what happened here is making more sense.

Initial sin- me being new to GFDL and python around two years ago, i re-code up a shell-script prototype that evolves into `regrid_xy`. Upon doing so, I mixed up `input` and `work` directories, and did not realize it, because `fregrid` does A LOT of file movement and it's easy to get mixed up on what it's doing. I rigged the unit tests *with that problem in place*, so this mixup gets tested and re-confirmed as "good" by the pipeline with every success. 

Then @singhd789 and @ceblanton encounter the tool, realize it needs edits, get it to what they want, and see the tests pass interactively/locally, but failing in the pipeline. The reason that failure was only visible in the pipeline, is that `test_success_tar_grid_spec_regrid_xy` that runs on GFDL/workstation, hides a failing condition that `test_success_no_tar_grid_spec_regrid_xy` would typically choke on. This proceeded to drive @ceblanton and @singhd789 crazy!!

Then I come back, thinking THEY've made the mistake, and slowly uncover what they learned but couldn't quite put into words because the local v. pipeline disparity adds confusion to the situation.  

Long story short is this- these unit tests no longer reflect the conditions we wish to test, so getting them to pass is misguided and we should mark them `xfail('NEED TO REFACTOR TESTS')`. A future refactor effort to re-engineer this tool and it's tests is being planned.

Issue #440, opened 2025-05-20
```fre pp checkout --branch``` fails in a confusing way


cwhitlock-NOAA:
This is a set of issues with fre pp checkout that came up in a thread in the MSD chat on 2025-05-16. 

- fre pp checkout --branch $branch will not update the code in ~/cylc-src/$e__$p__$t if there is already a branch $branch checked out. This behavior may have been implemented to keep from clobbering existing experiments with a changing codebase or to avoid unneded git operations, but now it means that you can fail to check out an updated development branch several times. 

Fix: add a --force-checkout or --force-update option that forces the code to pull the latest version from the git repo.

- if there is already an experiment under $e__$p__$t that's got a checked-out branch that does NOT match the branch $branch, the checkout fails. This wouldn't be terrible, but fre pp checkout grabs any nonzero exit status from the git command and helpfully tries to put in an error about what most likely went wrong - which, in this case, gives entirely the wrong impression.

Fix: allow --force-checkout or --force-update to change the branch, make a better error message (probably just pass on what we got from git)

Also I do not think that there are any tests for fre pp checkout aside from the dead simple "does the --help work" test.

cwhitlock-NOAA:
After talking with Ian on 05/21, it's probably also a good idea to allow an option to set which repo we're checking out from. For safety's sake, we might want to stick to only repos within NOAA *or* repos that are a fork of fre-workflows; there's a python package for git that might help with this.

PR #443, opened 2025-05-23
440.frepp.checkout.update

cwhitlock-NOAA:
## Describe your changes
Adding new args to fre pp checkout
  - --repo; allows for setting a new remote to clone from
  - --force-update; runs a git pull on a pre-existing clone to update it
  - --limit-checkout-size: limits the size of the fre-workflows checkout to the smallest we can get away with

Modifying the behavior around pre-existing checkouts
 - clearer error messages on why checking out a different branch under the same experiment fails (and how to fix)
 - --force-update allows for updating existing checkouts

Removing os.chdir() and replacing with git -C

This links to issue #440 

- [ ] I ran my code

### DIFF
--- a/fre/pp/checkout_script.py
+++ b/fre/pp/checkout_script.py
@@ -1,12 +1,20 @@
 '''
 Description: Checkout script which accounts for 4 different scenarios:
-1. branch not given, folder does not exist,
-2. branch given, folder does not exist,
-3. branch not given, folder exists,
-4. branch given and folder exists
+    default repo: 
+        default branch:
+            checkout folder does not exist
+            checkout folder exists
+            checkout folder exists, --force-update set to True
+        non-default branch:
+            ... (same 3 combos)
+    non-default repo
+        ... (same 6 combos)
+Non-default repo in combination with default branch is likely to give you 
+an error, but that's arguably a user problem
 '''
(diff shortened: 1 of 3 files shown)


cwhitlock-NOAA:
Question for review: should --force-checkout be able to check out a branch other than the one checked out previously? That is, should running 

```fre pp checkout --branch 51.var.filtering -e $experiment -p $platform -t $target
fre pp checkout --branch 51.bugfix.var.filtering -e $experiment -p $platform -t $target```

Give you an error?

I am currently inclined to say yes on the grounds that the "right" way to do this is to make sure no one triggers unexpected behavior by accident and either deleting your old directory in ~/cylc-src or setting a new $experiment both require a level of user intervention that you might not get if you're used to setting --force-update whenever you set your commands for fre pp checkout.

cwhitlock-NOAA:
Should we be trying to make the fre-workflows checkout as small as possible as our default behavior? We don't want anyone to be doing *development* from the directories in ~/cylc-src , so setting the git clone as small as possible in a way that you can't turn off from the command-line is arguably a good thing.

cwhitlock-NOAA:
We should probably accept both github and gitlab NOAA-GFDL urls.

Issue #410, opened 2025-04-29
Create remap-pp-components fre-cli tool


singhd789:
**Is your feature request related to a problem? Please describe.**
The task apps in fre-workflows can be re-written in python and made into fre-cli tools. The remap-pp-components script is already re-written and ready to be made into a fre-cli tool. 

**Describe the solution you'd like**
- Move the script from fre-workflows to `fre-cli/fre/app/`
- add click options/fix up script to work as fre-cli tool
- write/update remap-pp-components test

PR #463, opened 2025-06-11
Add remap-pp-components as fre-cli tool

singhd789:
#410 

- [x] I ran my code


fre/app/remap_pp_components/bin/data-lineage-remap-pp-components
singhd789, line 1: Note to self: where is this file used?

fre/app/helpers.py
```diff
@@ -0,0 +1,82 @@
+import os
+from pathlib import Path
+import yaml
+from contextlib import contextmanager
+
+# set up logging
+import logging
+logging.basicConfig(level=logging.INFO)
```
ilaflott, line 8: this should be removed pre-merge. calls to `logging.basicConfig` outside of the original one in `fre/__init__.py` should generally not happen!

fre/app/helpers.py
```diff
@@ -0,0 +1,82 @@
...
+fre_logger = logging.getLogger(__name__)
+
+def get_variables(yml, pp_comp):
```
ilaflott, line 11: function arg names are inconsistent with use in function body

fre/app/helpers.py
```diff
@@ -0,0 +1,82 @@
...
+
+    return src_vars
+
+## TEST get_variables ##
+#print(get_variables("/home/Dana.Singh/fre/get-vars-jinjafilter/Jinja2Filters/yaml_ex.yaml", "atmos_scalar atmos_scalar_test_vars atmos_scalar_test_vars_fail atmos_scalar_static_test_vars_fail"))
+
+## NOTE: For python 3.11 - this might be available already as contextlib.chdir()
+## Re-asses if our own contextmanager function is needed here
+@contextmanager
+def change_directory(new_path):
```
ilaflott, line 59: this isn't a bad idea... but, if this is something we wanna do (here and elsewhere) for transient `chdir`, we're gonna need some strong coverage/unit testing to make sure the behavior is reliable.

fre/app/remap_pp_components/remap_pp_components.py
```diff
@@ -0,0 +1,562 @@
...
+"""
+import os
+import subprocess
+import glob
+from pathlib import Path
+import logging
+import yaml
+from fre.app import helpers
+
+logging.basicConfig()
```
ilaflott, line 16: similar thought as above, remove `basicConfig` calls

fre/app/remap_pp_components/remap_pp_components.py
```diff
@@ -0,0 +1,562 @@
...
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
```
ilaflott, line 18: this should be removed pre-merge too. But, for the record, if it WERE left in (please don't), the resultant `logging` behavior is easier to work with, compared with leaving the extra `basicConfig` call in above.

ilaflott (commented):
WIP keep going!

fre/app/helpers.py
cwhitlock-NOAA: If it's a consistent function there's a couple other places in fre pp where it might be useful - or at least be a consistent design pattern. Then again, some of those argubaly could be tweaked a bit to avoid a chdir in the first place.

singhd789: Should unit testing for `helpers.py` go into a different PR though? I added a little test in relation to remap to make sure we finish where we started

ilaflott: Given that the function within  `helpers` was introduced with this PR, and the fact that we have all been bit too many times by an opaque error related to a non-transient `chdir`, I'm gonna be the bad guy that says, "yes it belongs in this PR". 

On the bright side, if that test was easy to write for `remap`, it probably looks very similar to what you'd write for `fre/app/helpers.py`.

singhd789: Ooooook, I suppoooooose you're right :laughing:

fre/app/helpers.py
```diff
@@ -0,0 +1,75 @@
...
+import logging
+fre_logger = logging.getLogger(__name__)
+
+def get_variables(yml, pp_comp):
+    """Retrieve any variables specified with active pp components from the yaml
+
+    :param yml: Already loaded yaml file
+    :type yml: str
+    :param pp_comp: List of active pp components
+    :type pp_comp: str
```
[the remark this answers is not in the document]
singhd789: good catch - list of strings

fre/app/remap_pp_components/README.md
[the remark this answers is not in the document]
singhd789: oooop, thanks. I did forget to update some things in there

cwhitlock-NOAA (changes requested):
Should this throw an error for situations where there is no overlap between pp_comp and component_info.get("type") for the components in postprocess components? I think that's likely to represent some kind of mistake in configuration - and if it's not a mistake in configuration, I'm not sure that the steps after this one for either split-netcdf or remap-pp-components are written to handle splitting 0 variables.

fre/app/helpers.py
```diff
@@ -0,0 +1,75 @@
...
+import logging
+fre_logger = logging.getLogger(__name__)
+
+def get_variables(yml, pp_comp):
+    """Retrieve any variables specified with active pp components from the yaml
+
+    :param yml: Already loaded yaml file
+    :type yml: str
+    :param pp_comp: List of active pp components
+    :type pp_comp: list of strings
```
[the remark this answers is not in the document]
singhd789: Oooook, so I have to fix something with this script, but the return in a dictionary of `{source1: [variable list], source2: [varlist], ...}`. 

I think your 2nd assumption is what I was following in my head. That is how the remap script is doing it. So, for each pp component that is passed, it find this dictionary of {src: [varlist]}. Would there be more of a need to have something like `{component: {src: [varlist]}, ...}`?

fre/app/helpers.py
```diff
@@ -0,0 +1,75 @@
...
+    :param pp_comp: List of active pp components
+    :type pp_comp: str
+    """
+    fre_logger.debug(f"Yaml file information: {yml}")
+    fre_logger.debug(f"PP components: {pp_comp}")
+
+    src_vars={}
+    for component_info in yml["postprocess"]["components"]:
+        # if component in yaml not an active pp component, skip
+        if component_info.get("type") not in pp_comp:
```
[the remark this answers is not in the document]
singhd789: I didn't throw an error because some components in the yaml have `postprocess_on: False`. We don't want to postprocess those, just skip because we don't care about them.

I also have changed this. Now pp_comp is only 1 component name being passed each time (in the workflow, I think it will be `$CYLC_TASK_PARAM_component`)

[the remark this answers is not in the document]
singhd789: not sure if that answered the question fully though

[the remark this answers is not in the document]
cwhitlock-NOAA: > I didn't throw an error because some components in the yaml have postprocess_on: False. We don't want to postprocess those, just skip because we don't care about them.

It sounds like from what you are saying that the parsing function is getting called with a pp_comp for each postprocessing component that's being processed in the yaml - which is why the list of postprocessing components that was originally passed to this function isn't the same thing as the list of all pp components in the yaml. I agree that not all components are going to be postprocessed, but I think the filtering step of looking for components that have postprocess_on: True takes place outside of this function (I'm going to double-check that tomorrow). 

I'd still argue that if pp_comp matches none of the components in the yaml this function either needs to throw an error or you need to state that the function can return an empty dictionary if the component and yaml don't match.

ilaflott:
might have to figure out how to turn down the verbosity on that bot...

fre/app/helpers.py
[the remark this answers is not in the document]
cwhitlock-NOAA: All right - could you add this to the documentation for the function? If you add a note about it being able to return an empty dictionary when there is not an overlap between the yaml and pp_comp, that would resolve the other open thread as well.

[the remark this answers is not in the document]
singhd789: ah nice catch! thank you, I have to add `returns`

cwhitlock-NOAA (approved):

Issue #497, opened 2025-06-23
docs: fre/pp


laurenchilutti:
This issue will log the task of documenting the files in fre/pp/ and docs/ related to fre/pp. Specifically looking to document:
classes
functions
function parameters

Add comments to code in general where needed.

Issue #394, opened 2025-04-10
Move tool: fre pp split-netcdf


cwhitlock-NOAA:
**Is your feature request related to a problem? Please describe.**
We would like to move away from keeping a lot of apps in fre-workflows. To do this, we've been steadily moving tools from fre-workflows to fre-cli. The next tool on this list is split-netcdf (currently split-netcdf-python); moving it here will make maintaining it and its dependencies easier and facilitate writing tests to integrate into our testing structure.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
This relates to the fre-workflows issue at: https://github.com/NOAA-GFDL/fre-workflows/issues/51

PR #511, opened 2025-06-24
docs: attempting header documentation for split-netcdf

cwhitlock-NOAA:
## Describe your changes
Updating headers in split-netcdf to reflect sphinx/readthedocs auto-documentation standards.

It also includes a second change - updating split-netcdf to use app/helpers.py::get_variables instead of split_netcdf::parse_yaml_for_varlist . It deletes some of the functions that had needed documentation in split-netcdf, removes some of the no-longer-needed unit tests, and updates the tests in test_split_netcdf_regex to use the app/helpers.py::get_variables function. 

Relates to issue #497 - fre pp documentation. I'm adding the documentation for split-netcdf. 

The second task relates to issues #410 (which added app/helpers.py) and the ongoing split-netcdf saga (#394 )

- [ ] I ran my code


laurenchilutti:
I think I have worked out the issues related to documentation not generating.  This was a very useful exercise in understanding Sphinx's requirements.  Here are the highlights that I have learned:

- You need a blank line between the parameter documentation `:paramter var1:` and any general comment text
- To continue text on the next line you need to indent to align with the start of the text above it
- You can't use `::` in some places that you have them.  I will comment directly on the code where this matters.

Here is an example:
```  
  Gets the encoding settings needed for xarray to write out the coordinates
  as expected
  we need the list of all vars (varnames) because that's how you get coords
  for the metadata vars (i.e. nv or bnds for time_bnds)

  :param dset: xarray Dataset object to query for info
  :type dset: xarray Dataset object
  :param vcoords: list of coordinate variables to write to file
  :type vcoords: list of strings
  :return: A dictionary where each key is a coordinate in the xarray Dataset and
           each value is a dictionary where the keys are the encoding information from
           the coordinate variable in the Dataset plus the units (if present)
  :rtype: dict

  .. note:: This code removes _FillValue from coordinates. CF-compliant files do not
            have _FillValue on coordinates, and xarray does not have a good way to get
            _FillValue from coordinates. Letting xarray set _FillValue for coordinates
            when coordinates *have* a _FillValue gets you wrong metadata, and bad metadata
            is worse than no metadata. Dropping the attribute if it's present seems to be
            the lesser of two evils.

```

fre/pp/tests/test_split_netcdf.py
```diff
@@ -93,13 +93,15 @@ def test_split_file_setup():
...
-        
+    :param workdir: subdir all operations are relative to
+    :type workdir: string
+    :param infile: netcdf file to split into single-var files
+    :type infile: string
+    :param outfiledir: directory to which to write the split netcdf files (new_all_ts_varlist, new_some_ts_varlist, new_none_ts_varlist)
+    :type outfiledir: string
+    :param varlist: comma-separated string specifying which variables to write ("all", some_ts_varlist, none_ts_varlist)
+    :type varlist: string
+    :type origdir: string ::
```
laurenchilutti, line 104: The sphinx auto doc builder doesnt like this `::`

fre/pp/split_netcdf_script.py
```diff
@@ -116,9 +123,12 @@ def split_file_xarray(infile, outfiledir, var_list='all'):
   variable name in the filename. 
   if var_list if specified, only the vars in var_list are written to file; 
   if no vars in the file match the vars in var_list, no files are written.
-  infile: input netcdf file
-  outfiledir: writeable directory to which to write netcdf files
-  var_list: python list of string variable names or a string "all"
+  :param infile: input netcdf file
```
laurenchilutti, line 126: you need to leave a blank line before you start using sphinx "field lists" - which is what the :param var1: technoiology is referred to as in the Sphinx world

fre/pp/split_netcdf_script.py
```diff
@@ -217,12 +229,15 @@ def set_coord_encoding(dset, vcoords):
...
-    only in relation with a variable
-  Note: this code removes _FillValue from coordinates. CF-compliant files do not
+  :param dset: xarray Dataset object to query for info
+  :type dset: xarray Dataset object
+  :param vcoords: list of coordinate variables to write to file
+  :type vcoords: list of strings
+  :return: A dictionary where each key is a coordinate in the xarray Dataset and 
+  each value is a dictionary where the keys are the encoding information from
+  the coordinate variable in the Dataset plus the units (if present)
+  :rtype: dict
```
laurenchilutti, line 239: It looks like we also need a blank loine between 239 and 240

fre/pp/split_netcdf_script.py
```diff
@@ -217,12 +229,15 @@ def set_coord_encoding(dset, vcoords):
...
+  ..note:: This code removes _FillValue from coordinates. CF-compliant files do not
```
laurenchilutti, line 240: And then lines 241 to 245 need to be indented to align with the "T" in "This code..."

fre/pp/split_netcdf_script.py
```diff
@@ -242,11 +257,19 @@ def set_coord_encoding(dset, vcoords):
 def set_var_encoding(dset, varnames):
   '''
   Gets the encoding settings needed for xarray to write out the variables
-  as expected
-  mostly addressed to time_bnds, because xarray can drop the units attribute:
+  as expected::
+  
+  mostly addressed to time_bnds, because xarray can drop the units attribute
     https://github.com/pydata/xarray/issues/8368
```
laurenchilutti, line 263: This line needs to have 2 less spaces (align with the text above it) so Sphinx stays happy

fre/pp/split_netcdf_script.py
```diff
@@ -262,24 +285,26 @@ def set_var_encoding(dset, varnames):
...
   Builds split  var filenames the way that fre expects them 
-  (and in a way that should work for any .nc file)
-  infile: string name of .nc file
-  varname: string to be added to input
-   Fre Input format:  date.component(.tileX).nc
-   Fre Output format: date.component(.tileX)var.nc
+  (and in a way that should work for any .nc file)::
+  
+   This is expected to work with files formed the following way
+    Fre Input format:  date.component(.tileX).nc
```
laurenchilutti, line 291: I would recommend makinig a bulleted list here.  Sphinx will let us do this by doing:
```
This is expected to work with files formed the following way:

- Fre Input format:  date.component(.tileX).nc
- Fre Output format: date.component.var(.tileX).nc

but it should also work on any file filename.nc
```

cwhitlock-NOAA: ...I aesthetically disagree with making this a bulleted list, but it probably is the easiest way to get that formatting.

fre/pp/split_netcdf_script.py
cwhitlock-NOAA: THANK YOU

fre/pp/split_netcdf_script.py
laurenchilutti: you dont have to go with the bulleted list! I encourage you to play around with the formatting!

cwhitlock-NOAA:
As of the latest update, the code coverage is fixed, the failing unit test from split_netcdf_regex is fixed, and I think that everything is good.

.gitignore
```diff
@@ -185,6 +185,7 @@ fre/tests/test_files/outdirfre/
 fre/tests/test_files/outdir/
 
 fre/app/generate_time_averages/tests/time_avg_test_files/
+fre/app/generate_time_averages/tests/test_data/
```
ilaflott, line 188: since we have `.cdl` files here, i wouldn't put this in the `.gitignore`

fre/app/helpers.py
```diff
@@ -22,8 +22,13 @@ def get_variables(yml: dict, pp_comp: str) -> dict:
     """
     fre_logger.debug(f"Yaml file information: {yml}")
     fre_logger.debug(f"PP component: {pp_comp}")
+    if not isinstance(yml, dict):
+        raise TypeError("yml should be of type dict, but was of type " + str(type(yml)))
 
     src_vars={}
+    print(yml)
```
ilaflott, line 29: no `print` if possible. 

as an aside though, in the case of `yaml` objects, `pprint` is standard library and does a lovely job of rendering a big thing readable.

fre/app/tests/test_helpers.py
```diff
@@ -40,6 +40,14 @@ def test_get_variables():
                 out2 == expected_dicts[1],
                 out3 == expected_dicts[2]])
 
+@pytest.mark.xfail(raises = TypeError)
```
ilaflott, line 43: i don't generally prefer `xfail` to test `exception` raising unless the testing-failure is both undesired and a consequence of active (desired) development.

if this failure is a temporary friction on the way to a solution in a future contribution, i'm good with this `xfail`. if this is testing what we WANT the code to do, then i'd recommend this get tested like: 

```    
with pytest.raises( ValueError ):
    helpers.get_variables(yml = YAML_EX, pp_comp = "test_param")
```

so that the test pops up with a green check instead of an ambiguous-yet-disconcerting yellow-x.

fre/app/tests/test_helpers.py
```diff
@@ -40,6 +40,14 @@ def test_get_variables():
                 out2 == expected_dicts[1],
                 out3 == expected_dicts[2]])
 
+@pytest.mark.xfail(raises = TypeError)                
+def test_get_variables_err():
+    """
+    Test get_variables() returns an error when given inappropriate input
+    """
+    out = helpers.get_variables(yml = YAML_EX, pp_comp = "test_param")
+    assert 0
```
ilaflott, line 49: `assert 0` here isn't doing anything, so i'd remove.

ilaflott (commented):
i'm actually still going over doc strings, but this feedback is outside the doc strings.

.gitignore
cwhitlock-NOAA: All right - I'll fix that

fre/app/tests/test_helpers.py
cwhitlock-NOAA: Okay - I'm going to restructure this test then.

fre/pp/split_netcdf_script.py
```diff
@@ -156,8 +176,9 @@ def split_file_xarray(infile, outfiledir, var_list='all'):
   var_exclude = list(set(var_patterns + [str(el) for el in var_shortvars] ))
   def matchlist(xstr):
     ''' checks a string for matches in a list of patterns
+    
         xstr: string to search for matches 
-        var_exclude: list of patterns defined in line 144'''
+        var_exclude: list of patterns defined in line 139'''
```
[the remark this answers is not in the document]
cwhitlock-NOAA: Okay: I fixed the reference here, but the regex doesn't generalize that way.

ilaflott (approved):
is good!

Issue #369, opened 2025-02-26
fmr/2025.01 module not loading mkmf


thomas-robinson:
**Describe the bug**
`fre make run-fremake -y ESM4.5v06_b08_cobv3_mekep_piC.yaml -p ncrc5.intel23 -t prod-openmp -npc -e`
Finishes with the error mkmf: Command not found.

**To Reproduce**
```
module load fre/2025.01
which mkmf
```

**Expected behavior**
mkmf should be loaded in the user's path when loading the module.

**Additional context**
This is on a gaea c5 log in node

Issue #282, opened 2024-12-03
Fre make --execute script fail on podman build does not stop the script from running


thomas-robinson:
**Describe the bug**
I was building a container using `fre make run-fremake ... --execute`.  My podman build failed in an expected way, but the script continued and a docker-archive file was made because I had previously build a container with the same tag.  The script should crash if the `podman build `is unsuccessful.

**To Reproduce**
`fre make run-fremake -y am5.yaml -p hpcme.2024 -t debug -npc --execute` with my yaml set up in /home/Thomas.Robinson/fre-examples_old/AM5/am5-compile on the AMD box before the mkmf changes go into effect.

**Expected behavior**
If any part of the build fails, the build should stop.

**Additional context**
podman containers will be overwritten if a build is successful.

singhd789:
Note: we should add `set -e` to the createContainer.sh` build script to stop if anything fails

PR #512, opened 2025-06-25
369.fre-cli mkmf dependency

singhd789:
## Describe your changes
- put mkmf submodule under `fre` 
- add mkmf path to PATH in `meta.yaml`
- could also help too: https://github.com/NOAA-GFDL/mkmf/blob/conda_package/meta.yaml (thanks Ian and Chris)

#369 

- [ ] I ran my code

### DIFF
--- a/.github/workflows/build_conda.yml
+++ b/.github/workflows/build_conda.yml
@@ -14,6 +14,9 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/noaa-gfdl/fre-cli:miniconda24.7.1_gcc14.2.0
+#    defaults:
+#      run:
+#        working-directory: /app
     steps:
       - name: Checkout Files
         uses: actions/checkout@v4
@@ -39,13 +42,6 @@ jobs:
           mkdir /app/fre-cli-tarball
           conda build --package-format tar.bz2 --output-folder /app/fre-cli-tarball .
 
-  upload:
-    runs-on: ubuntu-latest
-    needs: build
-    defaults:
-      run:
-        working-directory: /app
-    steps:
       - name: Upload fre-cli tarball
         uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/build_conda.yml
+++ b/.github/workflows/build_conda.yml
@@ -40,6 +40,7 @@ jobs:
 
   upload:
     runs-on: ubuntu-latest
+    needs: build
     defaults:
       run:
         working-directory: /app
--- a/.github/workflows/build_conda.yml
+++ b/.github/workflows/build_conda.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Add mkmf to GITHUB_PATH
         run: |
           # add mkmf to GITHUB_PATH
-          echo $PWD/mkmf/bin >> $GITHUB_PATH
+          echo $PWD/fre/mkmf/bin >> $GITHUB_PATH
 
       - name: Run Conda to Build
         run: |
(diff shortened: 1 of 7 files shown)


singhd789:
The create_test_conda_env pipeline failed before the build_conda pipeline - shouldn't it be testing/working in reverse? (building conda, then creating the test environment?)

meta.yaml
```diff
@@ -12,6 +12,7 @@ source:
 build:
   script:
     - {{ PYTHON }} -m pip install . -vv
+    - PATH=${PATH}:./mkmf/bin
```
ceblanton: The $PREFIX magic variable is the key here I think. This is the mechanism for how I did the mkmf conda build recipe- literally copying the bin/ stuff into the $PREFIX location.

https://gitlab.gfdl.noaa.gov/fre/hsm/-/blob/master/build.sh?ref_type=heads#L3

https://docs.conda.io/projects/conda-build/en/stable/resources/build-scripts.html

singhd789: Yes I think you're right! I was just trying to understand the process and just remembered that as well.

.github/workflows/build_conda.yml
```diff
@@ -20,10 +20,10 @@ jobs:
         with:
           submodules: 'recursive'
 
-      - name: Add mkmf to GITHUB_PATH
-        run: |
-          # add mkmf to GITHUB_PATH
-          echo $PWD/mkmf/bin >> $GITHUB_PATH
+#      - name: Add mkmf to GITHUB_PATH
```
ilaflott, line 23: clean up!

.github/workflows/publish_conda.yml
```diff
@@ -13,9 +13,9 @@ jobs:
         uses: actions/checkout@v4
         with:
           submodules: 'recursive'
-      - name: Add mkmf to PATH
-        run: |
-          echo $PWD/mkmf/bin >> $GITHUB_PATH
+#      - name: Add mkmf to PATH
```
ilaflott, line 16: clean up!

meta.yaml
```diff
@@ -50,7 +53,7 @@ test:
   source_files:
     - fre/
   files:
-    - mkmf/templates/linux-ubuntu-xenial-gnu.mk
+    - fre/mkmf/templates/linux-ubuntu-xenial-gnu.mk
```
ilaflott, line 56: should we just do all the templates?

ilaflott (changes requested):
basically there- clean up a touch and this can merge.

meta.yaml
singhd789: thought I'm not sure if we need aaaaaaall of the templates, idk. Will it make a big difference if we just did ` fre/mkmf/templates/*`? 🤔 interesting

singhd789:
Note for Dana:
- upload fre-cli tarball job succeeds even when tarball path was wrong --> make pipeline fail if this does not upload correctly?
-

ilaflott:
this should close both #282 and #369

.github/workflows/build_conda.yml
```diff
@@ -41,4 +36,14 @@ jobs:
...
+          mkdir -p /app/fre-cli-tarball
+          conda build --package-format tar.bz2 --output-folder /app/fre-cli-tarball .
+          ls /app/fre-cli-tarball
+
+## TO-DO: work this out
+      - name: Upload fre-cli tarball
+        uses: actions/upload-artifact@v4
+        with:
+          name: fre-cli-tarball
+          path: /app/fre-cli-tarball/noarch/fre-cli-2025.03-py_1.tar.bz2
```
ceblanton, line 48: Is there a way to avoid hard-coding the version here?

PR #517, opened 2025-06-30
docs: Update checkout_script.py

kiihne-noaa:

- [x] I ran my code


fre/pp/checkout_script.py
```diff
@@ -79,7 +93,7 @@ def checkout_template(experiment = None, platform = None, target = None, branch
         else:
             fre_logger.info(
                 f"ERROR: checkout exists ('{directory}/{workflow_name}') and does not match '{git_clone_branch_arg}'")
-            fre_logger.info(
+            fre_logger.info(
```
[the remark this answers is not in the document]
kiihne-noaa: removed

fre/pp/checkout_script.py
```diff
@@ -20,6 +20,18 @@
 def checkout_template(experiment = None, platform = None, target = None, branch = None):
     """
     Checkout the workflow template files from the repo
+
+    :param experiment: name of post processing experiment
+    :type experiment: str, default None
+    :param platform: which platform to use
+    :type c: str, default None
```
singhd789, line 27: should be `:type platform:`

Kiihne: Thanks for the catch. Should be ready for approval now

kiihne-noaa: So weird, I remember committing, but was not there. Fixed now.

fre/pp/checkout_script.py
```diff
@@ -20,6 +20,18 @@
...
     Checkout the workflow template files from the repo
+
+    :param experiment: name of post processing experiment
+    :type experiment: str, default None
+    :param platform: which platform to use
+    :type platform: str, default None
+    :param target: name of target (prod, debug, open-mp, repro)
+    :type target: str, default None
+    :param branch: which git branch to pull from
+    :type branch: str, default None
```
singhd789, line 31: I believe there is a default for this one.

fre/pp/checkout_script.py
```diff
@@ -20,6 +20,18 @@
...
+    :raises OSError: why checkout script was not able to be created
+    :return: directory with checkout script
+    :rtype: dir
```
singhd789, line 34: @laurenchilutti, is `rtype` supposed to be anything that has been specifically `returned` in the script or the general output/what was created from this script?

fre/pp/checkout_script.py
cwhitlock-NOAA: Is the default in another part of the workflow? We had a debate over this argument when the tool was being written - it was originally hard-coded to branch=main within the code, but given that we want to keep track of code via different branches in the future, branch=main is probably ok behavior for testing but not for production work.

singhd789: Ah wait, no default branch, but default tag! 
https://github.com/NOAA-GFDL/fre-cli/blob/main/fre/pp/checkout_script.py#L29

fre/pp/checkout_script.py
Kiihne: It is not specifically returned, but is what the script creates. Should I remove?

fre/pp/checkout_script.py
Kiihne: looks like the branch defaults to default tag of fre_ver if branch == None, so I will state fre_ver as the default branch

fre/pp/checkout_script.py
laurenchilutti: I would say it is specifically what would be returned.  If the script creates this directory, I would recommend updating the desciription to say something like `Create a directory and checkout the workflow template files from the repo` (that might not be exactly qhat to say, but I would recommend adding this to the description section and keep the return as none

fre/pp/checkout_script.py
ilaflott: pretty sure the default arg in the `python` function IS `None`... 

... we all see the `def checkout_template(experiment = None, platform = None, target = None, branch = None)` bit, right? 

what @singhd789 is pointing out here is probably worth a discussion on how default configuration is pulled in, and whether or not that should be happening is a conversation to be had elsewhere. 

regardless. the `docstring` docs the function, as it is written.

fre/pp/checkout_script.py
laurenchilutti: per our discussions in the fre rewrite meeting yesterday - you would have no return or rtype for this function (or any function that does not explicitly return a variable

fre/pp/checkout_script.py
```diff
@@ -20,6 +20,18 @@
...
+
+    :param experiment: name of post processing experiment
+    :type experiment: str, default None
+    :param platform: which platform to use
+    :type platform: str, default None
+    :param target: name of target (prod, debug, open-mp, repro)
+    :type target: str, default None
+    :param branch: which git branch to pull from
+    :type branch: str, default None
+    :raises OSError: why checkout script was not able to be created
```
laurenchilutti, line 32: It looks like this routine also raises 2 ValueErrors.  I think we should document them here too: you could do something like this: 
```
:raises ValueError: 

    - Description of error 1
    - Description of error 2
```

laurenchilutti (changes requested):

fre/pp/checkout_script.py
kiihne-noaa: implementing then resolving this issue

ilaflott (approved):
https://noaa-gfdl--517.org.readthedocs.build/projects/fre-cli/en/517/fre.pp.checkout_script.html

looks good to me!

laurenchilutti (approved):

Issue #521, opened 2025-07-02
WF Testing: make CI/CD `pylint` fail at a score threshold


ilaflott:
the current `pylint` score in the `main` branch is about [7.20](https://github.com/NOAA-GFDL/fre-cli/actions/runs/15979879969/job/45071556717#step:7:2441)

let's make this workflow step generally more useful for [`create_test_conda_env`](https://github.com/NOAA-GFDL/fre-cli/blob/6e5e9f01105eac3013f00bcadb0414909896907b/.github/workflows/create_test_conda_env.yml#L70), and the `build`/`publish_conda` workflows that run [`pylint` here](https://github.com/NOAA-GFDL/fre-cli/blob/6e5e9f01105eac3013f00bcadb0414909896907b/meta.yaml#L71)

- [ ] configure `pylint` to fail at a certain score-cut off and make sure that works somehow
- [x] set the cut-off score to be no-less than what is currently in `main`
- ongoing mission afterwards: configure `pylint` to be strict/permissive where we want

cwhitlock-NOAA:
Question: are we running this for newly-changed files, or for the repo as a whole?

cwhitlock-NOAA:
MiKyung suggested using https://github.com/NOAA-GFDL/pace as a reference for linting requirements, but they're using pre-commit and I'm having trouble tracking down configuration. Ask Frank about this later?

singhd789:
> Question: are we running this for newly-changed files, or for the repo as a whole?

Looks like it's for `fre/`

PR #523, opened 2025-07-03
adding check for pylint code threshold failure

cwhitlock-NOAA:

Issue #521 

- [ ] I ran my code


ilaflott (commented):
i think you might be able to just add a `--fail-under`, check `pylint --help | grep fail`

PR #527, opened 2025-07-07
docs: `fre app` generate time averages docstrings

kiihne-noaa:

- [x] I ran my code

### DIFF
--- a/fre/app/generate_time_averages/generate_time_averages.py
+++ b/fre/app/generate_time_averages/generate_time_averages.py
@@ -20,6 +20,8 @@ def generate_time_average(infile = None, outfile = None,
     :type unwgt: bool
     :param avg_type: time scale for climatology. Accepted variables vary based on pkg ('all','seas','month'), defaults to 'all'
     :type avg_type: str
+    :return: error message if requested package unkown, otherwise returns climatology
+    :rtype: int
     """
     if __debug__:
         fre_logger.info(locals()) #input argument details


fre/app/generate_time_averages/cdoTimeAverager.py
```diff
@@ -9,7 +9,17 @@ class inheriting from abstract base class timeAverager
...
-        ''' use cdo package routines via python bindings '''
+        """
+        use cdo package routines via python bindings        
+
+        :param self: 
+        :param infile: path to history file, or list of paths
+        :type infile: str, list
+        :param outfile: path to where output file should be stored
+        :type outfile: str
+        :param self: time scale for climatology. Accepted variables vary based on pkg ('all','seas','month')
```
ceblanton, line 20: I'm not sure this object attribute should be in the docstring, but it does look useful.

fre/app/generate_time_averages/cdoTimeAverager.py
```diff
@@ -9,7 +9,17 @@ class inheriting from abstract base class timeAverager
     '''
 
     def generate_timavg(self, infile=None, outfile=None):
-        ''' use cdo package routines via python bindings '''
+        """
```
ceblanton, line 12: I think we agreed to do the type hints as part of this doc update. Can you add those?

ceblanton (dismissed):
Please add type hints to the method definitions, and let's discuss whether to mention the object attributes.

fre/app/generate_time_averages/cdoTimeAverager.py
```diff
@@ -9,7 +9,17 @@ class inheriting from abstract base class timeAverager
...
+        use cdo package routines via python bindings        
+
+        :param self:
```
laurenchilutti, line 15: I think the recommendation will be to document this as `:param self: This is an instance of the class cdoTimeAverager` and I would not define a type for it.

So you can keep self as the first parameter and document it here, and then delete lines 20 and 21 I think.

fre/app/generate_time_averages/cdoTimeAverager.py
```diff
@@ -9,7 +9,17 @@ class inheriting from abstract base class timeAverager
...
+        use cdo package routines via python bindings        
+
+        :param self: 
+        :param infile: path to history file, or list of paths
+        :type infile: str, list
+        :param outfile: path to where output file should be stored
+        :type outfile: str
+        :param self: time scale for climatology. Accepted variables vary based on pkg ('all','seas','month') 
+        :type self: str, defaults to 'all'. Inherted from generate_time_averages.py
+        """
```
laurenchilutti, line 22: I think we should also document the returns (0 and 1).  Im thinking: 
```
:return: 1 if the instance variable self.avg_typ is unsupported, 0 if function has a clean exit
:rtype: int
```

fre/app/generate_time_averages/frenctoolsTimeAverager.py
```diff
@@ -17,7 +17,17 @@ class inheriting from abstract base class timeAverager
     '''
 
     def generate_timavg(self, infile=None, outfile=None):
-        ''' use fre-nctool's CLI timavg.csh with subprocess call '''
+        """
+        use fre-nctool's CLI timavg.csh with subprocess call
+
+        :param self:
```
laurenchilutti, line 23: I think the recommendation will be to document this as `:param self: This is an instance of the class frenctoolsTimeAverager` and I would not define a type for it.

So you can keep self as the first parameter and document it here, and then delete lines 28 and 29 I think.

fre/app/generate_time_averages/frenctoolsTimeAverager.py
```diff
@@ -17,7 +17,17 @@ class inheriting from abstract base class timeAverager
...
+        use fre-nctool's CLI timavg.csh with subprocess call
+
+        :param self: 
+        :param infile: path to history file, or list of paths
+        :type infile: str, list
+        :param outfile: path to where output file should be stored
+        :type outfile: str
+        :param self: time scale for climatology. Accepted variables vary based on pkg ('all','seas','month') 
+        :type self: str, defaults to 'all'. Inherted from generate_time_averages.py
+        """
```
laurenchilutti, line 30: I see some ValueErros raised in this function.  My recommendation is to include that here:
```
:raises ValueError:

    - instance variable self.avg_type not supported
    - No infile specified
    - Cannot find timavg.csh (likely the user is not in an environment with frenctools installed)
    - timavgcsh command is not properly executed
```

I also see we return an exit code:
```
:return: 1 if timavg.csh command is not properly executed, and 0 if function has a clean exit
:rtype: int
```

fre/app/generate_time_averages/frepytoolsTimeAverager.py
```diff
@@ -14,8 +14,19 @@ class inheriting from abstract base class timeAverager
...
 
     def generate_timavg(self, infile=None, outfile=None):
-        ''' frepytools approach in a python-native manner.
-        deliberately avoids pre-packaged routines. '''
+
+        """
+        frepytools approach in a python-native manner.
+        deliberately avoids pre-packaged routines.
+
+        :param self:
```
laurenchilutti, line 22: same comment on the param self as the previous 2 files, noting that you shouldin this file refer to it as an instance of frepytoolsTimeAverager class

fre/app/generate_time_averages/frepytoolsTimeAverager.py
```diff
@@ -14,8 +14,19 @@ class inheriting from abstract base class timeAverager
...
+        deliberately avoids pre-packaged routines.
+
+        :param self: 
+        :param infile: path to history file, or list of paths
+        :type infile: str, list
+        :param outfile: path to where output file should be stored
+        :type outfile: str
+        :param self: time scale for climatology. Accepted variables vary based on pkg ('all','seas','month') 
+        :type self: str, defaults to 'all'. Inherted from generate_time_averages.py
+        """
```
laurenchilutti, line 29: Add the returns 1 and 0 similarly to the other 2 files.  But the description of a return value of 1 will be unique to describe the cases where we would return 1 (lines 39 and 68)

fre/app/generate_time_averages/generate_time_averages.py
```diff
@@ -5,7 +5,22 @@
 def generate_time_average(infile = None, outfile = None,
                           pkg = None, var = None, unwgt = False,
                           avg_type = None):
-    ''' steering function to various averaging functions above'''
+    """
+    steering function to various averaging functions above
+    
+    :param infile: path to history file, or list of paths
```
laurenchilutti, line 11: I recommend specifiying in the param description for each parameter that the default is None. Not in the type.

fre/app/generate_time_averages/generate_time_averages.py
```diff
@@ -5,7 +5,22 @@
...
+    :type infile: str, list
+    :param outfile: path to where output file should be stored
+    :type outfile: str
+    :param pkg: which package to use to calculate climatology (cdo, fre-nctools, fre-python-tools)
+    :type pkg: str
+    :param var: not currently supported
+    :type var: defaults to none
+    :param unwgt: wether or not to weight the data
+    :type unwgt: True or False, default false
```
laurenchilutti, line 20: I think we would say the type is `bool`, move the default to the param description (line above)

fre/app/generate_time_averages/generate_time_averages.py
```diff
@@ -5,7 +5,22 @@
...
+    steering function to various averaging functions above
+    
+    :param infile: path to history file, or list of paths
+    :type infile: str, list
+    :param outfile: path to where output file should be stored
+    :type outfile: str
+    :param pkg: which package to use to calculate climatology (cdo, fre-nctools, fre-python-tools)
+    :type pkg: str
+    :param var: not currently supported
+    :type var: defaults to none
```
laurenchilutti, line 18: I think you can delete this line since we don't support it.  If you do keep this line make sure we put in an actual type like int, str, etc.

fre/app/generate_time_averages/generate_time_averages.py
```diff
@@ -5,7 +5,22 @@
...
+    :param unwgt: wether or not to weight the data
+    :type unwgt: True or False, default false
+    :param avg_type: time scale for climatology. Accepted variables vary based on pkg ('all','seas','month') 
+    :type avg_type: str, defaults to 'all'
```
laurenchilutti, line 22: move this "defaults to 'all'" to the param description on the line above.

laurenchilutti (changes requested):

fre/app/generate_time_averages/generate_time_averages.py
kiihne-noaa: Done across all three files

kiihne-noaa:
@laurenchilutti @ilaflott @ceblanton I have completed all your comments. There was a lot of useful feedback to incorporate. Please let me know with a quick second pass if there are any other adjustments to be made.

fre/app/generate_time_averages/generate_time_averages.py
ilaflott: is specifying `None` as default really that needed in the description? the function `args` [here](https://noaa-gfdl--527.org.readthedocs.build/projects/fre-cli/en/527/fre.app.generate_time_averages.cdoTimeAverager.html?readthedocs-diff=true&readthedocs-diff-chunk=1#module-fre.app.generate_time_averages.cdoTimeAverager) are displayed pretty clearly automatically

ilaflott (approved):

laurenchilutti (approved):

Issue #528, opened 2025-07-07
`cylc scan [experiment]` not found


singhd789:
**Describe the bug**
`subprocess.run` prefers a sequence of arguments rather than a string. If a single string is used, `shell=True` has to be specified or the string cannot specify arguments (https://docs.python.org/3/library/subprocess.html#frequently-used-arguments). 

When trying to use `fre pp run` currently, I kept seeing the error:
`FileNotFoundError: [Errno 2] No such file or directory: 'cylc scan --name c96L65_am5f8d6r3_amip__ppan__debug'`

**To Reproduce**
Going through regular fre pp steps: 
```
fre pp checkout ...
fre pp configure-yaml ...
fre pp install ...
fre pp run ...
```

**Expected behavior**
The experiment should kick off with the `run` step but it stops and spits out that `File not found` message

**Additional context**
Simple solution:  edit [line 24](https://github.com/NOAA-GFDL/fre-cli/blob/2d265b285102a73346407e5e4c560d6ea3d40b81/fre/pp/run_script.py#L24)

`result = subprocess.run( first_cmd, capture_output = True ).stdout.decode('utf-8')` -->  `result = subprocess.run(['cylc', 'scan', '--name', f"^{name}$"], capture_output = True ).stdout.decode('utf-8')`

ceblanton:
Thanks Dana.. (How did it ever work.. ahhh!)

PR #529, opened 2025-07-07
Fix "cylc scan --name [experiment] not found"

singhd789:
## Describe your changes
The issue is related to the `cylc scan` call in the `fre pp run` step. When going through the fre pp steps, `fre pp run` would spit out and error `cylc scan --name [experiment] not found`. I think this is related to the format when using `subprocess.run`. According to this doc, https://docs.python.org/3/library/subprocess.html#frequently-used-arguments,  the arguments in the call can be a string or a sequence of arguments, but the sequence is preferred. When I changed it back to a sequence, I did not see this `cylc scan` error. 

- imported `make_workflow_name` to create name of workflow as other `fre pp` scripts do
- wrote out arguments in subprocess run call 
- to-do?: create some kind of test for this

#528 

- [x] I ran my code


ilaflott:
Can we get a better description of the original problem? i.e. what was the call and what was the actual output?

singhd789:
> Member

updated the description from what I could remember from last week!

singhd789:
> Can we get a better description of the original problem? i.e. what was the call and what was the actual output?

updated the description from what I could remember from last week!

ilaflott (approved):

PR #591, opened 2025-08-13
remove uncovered `__name__=="__main__"` entrypoints 

ilaflott:
not needed for package functionality as far as i am aware. if any are needed, we should test the use case! otherwise, they will remain uncovered / unused and we should remove them.

my other motivation is the reduce the size of #429 by introducing certain non-CMOR motivated changes here


ceblanton (approved):
Thanks Ian, this cleanup is 100% safe, looks great.

Issue #634, opened 2025-09-05
fre pp split-netcdf-wrapper yaml parsing function


cwhitlock-NOAA:
fre pp split-netcdf-wrapper currently relies on the same parsing function as remap-pp-components, which takes in a component (type in the yamlfile) and a history_source underneath it and grabs the list of variables associated with that history_source. 

However, if you look at flow.cylc in fre-workflows - where we intend to use fre pp split-netcdf-wrapper - we aren't cylcing over <component> (type in the yamlfile) during the split-netcdf steps. We're cylcing over <regrid> and its variants, which are a flat list of all history_sources in the input yamlfile.

This means that split-netcdf-wrapper can't use the same parsing function as remap-pp-components - at least not  yet. If we get around to that big refactor of flow.cylc it would probably make sense to make the split-netcdf tasks loop on type as well as history_source, but the appropriate intermediate step is for split-netcdf-wrapper to parse yamlfiles differently than remap-pp-components.

reference yamlfile: https://github.com/NOAA-GFDL/fre-cli/blob/main/fre/app/remap_pp_components/tests/test-data/yaml_ex.yaml

cwhitlock-NOAA:
I still think that the feature for this is relevant even if the branch that addressed it got deleted in a cleanup.

PR #635, opened 2025-09-05
new function for parsing yaml when only a hist_source is given

cwhitlock-NOAA:
## Describe your changes
split-netcdf-wrapper can't use the same function for parsing yamlfiles that remap-pp-components does, even if they're pulling the same information from the yamlfile (the variable list associated with a history_file). In flow.cylc, remap-pp-components is cycling over <component>, but split-netcdf is cylcing over <regrid/native> (a flat list of history_files). 

It's inelegant and goes against the structure of the doc, but the fact that the history_files are unique means that we can get the information we need for split-netcdf with just a history_file.

This relates to issue #634 and the ongoing set of issues around split-netcdf (fre-workflows pull request #100). 

Reference yamlfile: https://github.com/NOAA-GFDL/fre-cli/blob/main/fre/app/remap_pp_components/tests/test-data/yaml_ex.yaml

- [x] I ran my code


cwhitlock-NOAA:
For the record - I'm aiming to get this included in the 2026 release, not the 2025 release. This is not an urgent task (aside from the fact that it's part of something that's been dangling for four months).

fre/app/helpers.py
```diff
@@ -61,6 +60,57 @@ def get_variables(yml: dict, pp_comp: str) -> dict:
...
+        raise TypeError("yml should be of type dict, but was of type " + str(type(yml)))
+    is_found = False
+    src_vars = []
+    for component_info in yml["postprocess"]["components"]:
+        #it is assumed that hist_src is pre-filtered for active/inactive pp components
+        src_info = component_info.get("sources")
+        #timeseries case
+        for src in src_info:
+            if src['history_file'] == hist_src:
+                is_found = True
```
ilaflott, line 92: nit, i think this reads a bit nicer like:
```
is_found = bool( src['history_file'] == hist_src )
if is_found:
    ...
```

fre/app/tests/test_helpers.py
```diff
@@ -39,15 +45,45 @@ def test_get_variables():
...
         out = helpers.get_variables(yml = YAML_EX, pp_comp = "test_param")
     assert execinfo.type is TypeError
 
+@pytest.mark.parametrize("hist_src,var_list",
+                          [
+                          pytest.param("atmos_month", "all", id="tseries-all"),
+                          pytest.param("atmos_static_scalar", "all", id="static-all"),
+                          pytest.param("atmos_scalar_test_vars",["co2mass"], id="tseries-varlist"),
+                          pytest.param("atmos_static_scalar_test_vars", ["bk"] , id="static-varlist"),
+                          pytest.param("atmos_daily", "all", id="fail-nohist_src", marks=pytest.mark.xfail())
```
ilaflott, line 63: does the `marks=pytest.mark.xfail` bit in `parameterize` mean any test routine calling the parameters will `xfail`?

fre/app/tests/test_helpers.py
```diff
@@ -39,15 +45,45 @@ def test_get_variables():
...
+    assert new_var_list == var_list
+    
+def test_get_var_hist_src_load_wrong_type():
+    '''
+    Checks that we get the right error raised when we try to read an unloaded yaml
+    for get_variables_hist_src
+    '''
+    with pytest.raises( TypeError ) as execinfo2:
+        out1 = helpers.get_variables_hist_src(YAML_EX, 'atmos_scalar_test_vars')
+    assert execinfo2.type is TypeError
```
ilaflott, line 85: i think this line might be uneeded? the `with pytest.raises(TypeError)` bit should cover it, i believe.

fre/pp/frepp.py
```diff
@@ -178,23 +176,22 @@ def histval(history,date_string,warn):
               help="Whether to search subdirs underneath $inputdir for netcdf files. Defaults to false. This option is used in flow.cylc when regridding.")
 @click.option('--split-all-vars', '-a', is_flag=True, default=False,
               help="Whether to ignore other config options and split all vars in the file. Defaults to false. Conflicts with -c, -s and -y options.")
-def split_netcdf_wrapper(inputdir, outputdir, component, history_source, use_subdirs, yamlfile, split_all_vars):
+def split_netcdf_wrapper(inputdir, outputdir, history_source, use_subdirs, yamlfile, split_all_vars):
     ''' Splits all netcdf files matching the pattern specified by $history_source in $inputdir
         into files with a single data variable written to $outputdir. If $yamlfile contains
-        variable filtering settings under $component, only those variables specified will
+        variable filtering settings under $history_source, only those variables specified will
```
ilaflott, line 182: nit, omit the `$`

ilaflott (commented):
looks pretty good, a Q or two and a nit here/there but probably mergeable

PR #656, opened 2025-09-23
Replace fre-nctools conda dependency with building it from source

ceblanton:
## Describe your changes
The noaa-gfdl channel fre-nctools package was built on the GFDL Intel-based workstations, and so does not work with the AMD-based container. But if we build fre-nctools as part of the fre-cli conda build process, then it will by definition be compiled on the correct architecture.

- [ ] I ran my code

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -29,6 +29,8 @@ channels:
     - noaa-gfdl
 
 requirements:
+  build:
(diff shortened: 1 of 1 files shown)


ilaflott:
this, we agreed, is not likely to work. goes against all `conda` packaging wisdom

Issue #659, opened 2025-10-01
`fre make all` refactor/clean up


singhd789:
**Is your feature request related to a problem? Please describe.**
The `fre make all` subtool should really be a wrapper calling the other fre make tools (similar to `fre pp all`). This way it would be more organized, cleaner, and would be easier to fix if needed. (currently if a fre make subtool needs to be fixed, the same fix needs to be done for fre make run - we don't like the duplication!)

**Describe the solution you'd like**
Make the `fre make all` tool a wrapper calling the rest of the `fre make` subtools

PR #660, opened 2025-10-01
 #659 reorganize `fre make all`

singhd789:

- [ ] I ran my code

### DIFF
--- a/fre/make/run_fremake_script.py
+++ b/fre/make/run_fremake_script.py
@@ -3,16 +3,11 @@ For a bare-metal build: Creates and runs the checkout script to check out source
 
 For a container build: Creates the checkout script and makefile, and creates and runs a dockerfile to generate a singularity image file.
 '''
-
-import os
 import logging
-
-from multiprocessing.dummy import Pool
-from pathlib import Path
-
-import subprocess
-import fre.yamltools.combine_yamls_script as cy
-from typing import Optional
+from fre.make.create_checkout_script import checkout_create 
+from fre.make.create_makefile_script import 
+from fre.make.create_compile_script import
+from fre.make.create_docker_script import
 from .gfdlfremake import (
     targetfre, varsfre, yamlfre, checkout,
     makefilefre, buildDocker, buildBaremetal )
@@ -44,35 +39,20 @@ def fremake_run(yamlfile:str, platform:str, target:str, nparallel: int = 1, njob
     :type execute: bool
     :param verbose: Increase verbosity output
     :type verbose: bool
-    :raise ValueError:
-        - Error if platform does not exist in platforms yaml configuration 
-        - Error if the mkmf template defined in platforms yaml does not exist
-
-    .. note:: This script will eventually be a wrapper for the other fre make tools
     """
-    yml = yamlfile
+#    if verbose:
+#        fre_logger.setLevel(level = logging.DEBUG)
+#    else:
+#        fre_logger.setLevel(level = logging.INFO)
+
+    # Define variables
     name = yamlfile.split(".")[0]
-    nparallel = nparallel
-    jobs = str(njobs)
-    pcheck = no_parallel_checkout
-
-    if pcheck:
-        pc = ""
-    else:
-        pc = " &"
-
-    if verbose:
-        fre_logger.setLevel(level = logging.DEBUG)
-    else:
-        fre_logger.setLevel(level = logging.INFO)
-
-    #### Main
-    srcDir="src"
-    baremetalRun = False # This is needed if there are no bare metal runs
-
-    ## Split and store the platforms and targets in a list
     plist = platform
     tlist = target
+
+    ## Error checking the targets
+    for target_name in tlist:
+        target = targetfre.fretarget(target_name)
 
     # Combine model, compile, and platform yamls
     full_combined = cy.consolidate_yamls(yamlfile=yml,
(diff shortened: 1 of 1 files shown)


fre/make/create_docker_script.py
```diff
@@ -77,12 +77,20 @@ def dockerfile_create(yamlfile:str, platform:str, target:str, execute: Optional[
             image=modelYaml.platforms.getContainerImage(platformName)
             stage2image = modelYaml.platforms.getContainer2base(platformName)
             tmpDir = "tmp/"+platformName
+
+            # check if mkTemplate has a / indicating it is a path
+            # if its not, prepend the template name with the mkmf submodule directory
+            if "/" not in platform["mkTemplate"]:
+                templatePath = platform["modelRoot"]+"/mkmf/templates/"+platform["mkTemplate"]
+            else:
+                templatePath = platform["mkTemplate"]
```
ilaflott: copy/paste code --> some kind of function ?

fre/make/make_helpers.py
```diff
@@ -0,0 +1,46 @@
+''' this holds functions used across various parts of fre/make subtools '''
+
+import logging
+from pathlib import Path
+from typing import Optional
+
+# set logger level to INFO
+fre_logger = logging.getLogger(__name__)
+former_log_level = fre_logger.level
+fre_logger.setLevel(logging.INFO)
```
ilaflott: `fre_logger` isn't used here- perhaps we need some calls. But really why I'm here, is we should only set the `level` to `INFO` for the lines below that you want, and no more than those lines. It's to give the user as fine-grained control over the output as possible.

fre/make/create_docker_script.py
```diff
@@ -11,6 +11,7 @@
 
 import fre.yamltools.combine_yamls_script as cy
 from typing import Optional
+import fre.make.make_helpers as mh
```
ilaflott: nit/taste, if you're only gonna call the one function from `make_helpers` once, i'd make this `from fre.make.make_helpers import the_function`

fre/make/create_compile_script.py
```diff
@@ -10,6 +10,7 @@
 
 import fre.yamltools.combine_yamls_script as cy
 from typing import Optional
+import fre.make.make_helpers as mh
```
ilaflott: same as other line

fre/make/tests/compilation/test_run_fremake_builds.py
```diff
@@ -114,4 +114,29 @@ def test_run_fremake_cleanup():
             print(tp + " not found for deletion. Something may have gone wrong elsewhere.")
     tp_remove = [not Path(el).exists() for el in test_paths]
     assert all(tp_remove)
-    
+
+def test_run_fremake_container_build_fail():
+    ''' check createContainer script would fail and exit if one step failed '''
```
ilaflott: nit/taste, i prefer the `'''` to be on their own lines

fre/make/tests/compilation/test_run_fremake_builds.py
```diff
@@ -114,4 +114,29 @@ def test_run_fremake_cleanup():
...
+    if Path(f"{currPath}/createContainer.sh").exists():
+        os.remove(Path(f"{currPath}/createContainer.sh"))
```
ilaflott: i don't thiiiiiink `os.remove` needs a `Path`, the `str` should suffice and it will be easier to read

fre/make/tests/compilation/test_run_fremake_builds.py
```diff
@@ -114,4 +114,29 @@ def test_run_fremake_cleanup():
...
+        no_format_transfer=False, execute=False, verbose=VERBOSE)
+    assert Path(f"{currPath}/createContainer.sh").exists()
+
+    # Alter script to fail
+    with open(Path(f"{currPath}/createContainer.sh"), "r") as f:
+        lines = f.readlines()
+
+    with open(Path(f"{currPath}/createContainer.sh"), "w") as f:
+        for line in lines:
+            f.write(line.replace("Dockerfile", "Dockerfile-wrong"))
```
ilaflott: these can be combined- there is no writing without reading. There can however, be reading without writing.

fre/make/run_fremake_script.py
```diff
@@ -86,168 +68,38 @@ def fremake_run(yamlfile:str, platform:str, target:str, nparallel: int = 1, njob
...
+            raise ValueError (f"{platform_name} does not exist in platforms.yaml")
+
+        platform_info = model_yaml.platforms.getPlatformFromName(platform_name)
+
+        if not platform_info["container"]:
+            bm_platforms = bm_platforms + (platform_name,)
+        else:
+            container_platforms = container_platforms + (platform_name,)
+
+    if bm_platforms:
```
ilaflott: you wanna double check (I always forget) how `if <empty collection thing> ` (in this case, `tuple`) is gonna act. It's not always the way one expects

fre/make/gfdlfremake/buildDocker.py
```diff
@@ -235,7 +235,7 @@ def createBuildScript(self, platform, skip_format_transfer):
         registry_tag = self.e.lower()
         platform_tag = self.target.gettargetName().lower()
 
-        self.userScript = ["#!/bin/bash\n"]
+        self.userScript = ["#!/bin/bash\n", "set -ex\n"]
```
ilaflott: this probably is out of scope for this PR, but, given that `writelines` handles the writing, could it possibly append a newline/`\n` to each string in the list? then we could take out all the `\n` everywhere else

fre/make/make_helpers.py
singhd789: oooop, that's what copy and paste from another script get me

fre/make/run_fremake_script.py
singhd789: So is a bit right before this that basically checks for the existence of the platforms passed (https://github.com/NOAA-GFDL/fre-cli/blob/659.fre-make-all-reorganize/fre/make/run_fremake_script.py#L86-L87)

So if it doesn't exist, it will error out. If it does exist, it should continue to be appended to those variables). So reeeeeaally, it should never be empty if a correct platform is passed. Should I alter the if statement somehow or still add in another check you think?

fre/make/tests/compilation/test_run_fremake_builds.py
singhd789: Hmmmm, I also usually do that but wanted to be consistent with what was in the test file already (with the other tests)

rem1776 (approved):
lgtm!

PR #665, opened 2025-10-17
`fre app mask-atmos-plevel` edits for `fre-workflows` functionality

ilaflott:
- minor adjustments to the execution logic of `fre app mask-atmos-plevel` in light of workflow testing at https://github.com/NOAA-GFDL/fre-workflows/pull/64 
- more testing coverage, use-cases and exceptions
- two new cdl files... sigh
- made `pytest` fixtures more flexible in mask-atmos-plevel tests, letting them be called individually rather than as a monolith
- many many more `logging` statement calls for insight into whats actually being executed
- CLI oriented tests for `mask-atmos-plevel`
- testing code for `mask-atmos-plevel` no longer contributes to coverage measurements
- successful [workflows](https://github.com/NOAA-GFDL/fre-workflows/actions/workflows/test_cloud_runner.yml?query=branch%3Aatmos-pressure-masking+) with the `ppp` container and using this branch
- makes a `logging` level change transient in an adjacent submodule as a sidequest 

- [x] I ran my code

### DIFF
--- a/fre/app/mask_atmos_plevel/tests/test_mask_atmos_plevel.py
+++ b/fre/app/mask_atmos_plevel/tests/test_mask_atmos_plevel.py
@@ -211,7 +211,7 @@ def test_mask_atmos_plevel_no_ps_error(tmp_input, tmp_path): # pylint: disable=r
     assert not tmp_output.exists()
 
 
-def test_mask_atmos_plevel_nops_warn(tmp_input, tmp_path): # pylint: disable=redefined-outer-name
+def test_mask_atmos_plevel_no_ps_warn(tmp_input, tmp_path): # pylint: disable=redefined-outer-name
     """
     if the input file exists, but the ps file does not, then no-op gracefully
     without raising an exception or throwing an error while warn_no_ps = True


ilaflott:
this is being additionally tested with https://github.com/NOAA-GFDL/fre-workflows/pull/64

ilaflott:
This seems to have the functionality required by the fre-workflows PR. It needs some testing coverage however

fre/app/freapp.py
```diff
@@ -119,9 +119,13 @@ def regrid(yamlfile, input_dir, output_dir, work_dir,
               help = "Input NetCDF file containing surface pressure (ps)",
               type = str,
               required = True)
-def mask_atmos_plevel(infile, psfile, outfile):
+@click.option("-wnps", "--warn-no-ps",
```
singhd789: Could this be some kind of check in the actual tool? What's the benefit of having it be an actual flag here and where/in what case would it be used?

fre/app/mask_atmos_plevel/mask_atmos_plevel.py
```diff
@@ -14,55 +23,83 @@ def mask_atmos_plevel_subtool(infile: str, psfile: str, outfile: str) -> None:
...
         if 'pressure_mask' in ds_in[var].attrs:
             if ds_in[var].attrs['pressure_mask'].lower() == 'false':
                 ds_out[var] = mask_field_above_surface_pressure(ds_in, var, ds_ps)
                 ds_out[var].attrs['pressure_mask'] = "True"
-                fre_logger.info(f"Finished processing '{var}' and set 'pressure_mask' to True")
+                fre_logger.info("Finished processing %s, pressure_mask is True", var)
             else:
-                fre_logger.debug(f"Not processing '{var}' (because 'pressure_mask' is not False)")
+                fre_logger.debug("Not processing %s (because 'pressure_mask' is not False)", var)
+        elif '_unmsk' in var:
```
singhd789: just some clarification for me - what's `pressure_mask` vs `_unmsk`? A comment could be helpful

fre/app/mask_atmos_plevel/mask_atmos_plevel.py
```diff
@@ -14,55 +23,83 @@ def mask_atmos_plevel_subtool(infile: str, psfile: str, outfile: str) -> None:
...
+            fre_logger.debug('setting pressure_mask attribute of %s to True', masked_var)
+            ds_out[masked_var].attrs['pressure_mask'] = "True"
+
+            fre_logger.info("Finished processing %s, wrote %s, pressure_mask is True", var, masked_var)
         else:
-                fre_logger.debug(f"Not processing '{var}' (because it does not have 'pressure_mask'")
+            fre_logger.debug("Not processing %s, it does not have pressure_mask", var)
 
-    # Write the output file if anything was done
+    fre_logger.info('Write the output file if anything was done')
```
singhd789: What do you mean by "if anything was done"? Could be more clear

fre/app/mask_atmos_plevel/mask_atmos_plevel.py
```diff
@@ -78,34 +115,35 @@ def mask_field_above_surface_pressure(ds: xr.Dataset, var: str, ds_ps: xr.Datase
     .. note:: Missing values are set to 1.0e20.
     """
 
-    # retrieve the pressure coordinate variable
+    fre_logger.info('retrieve the pressure coordinate variable')
```
singhd789: maybe this is implied but would be good to throw in there - I believe plev is the vertical coordinate right?

I would just put that in there --> `fre_logger.info('retrieve the vertical pressure coordinate variable, plev')`

fre/app/mask_atmos_plevel/tests/test_mask_atmos_plevel.py
```diff
@@ -65,3 +166,110 @@ def test_mask_atmos_plevel(create_input_files):
...
+    """
+    tmp_output = Path(tmp_path / "output.nc")
+
+    # remove the ps file
+    tmp_ps.unlink()
+    assert not tmp_ps.exists()
+
+    mask_atmos_plevel.mask_atmos_plevel_subtool(infile = tmp_input,
+                                                psfile = tmp_ps,
+                                                outfile = tmp_output)
```
singhd789: to be more clear, even though it is the default already, I would throw in the flag created (if kept) - `warn_no_ps = False` like in the next test

fre/app/mask_atmos_plevel/tests/test_mask_atmos_plevel.py
```diff
@@ -65,3 +166,110 @@ def test_mask_atmos_plevel(create_input_files):
...
+                                                warn_no_ps = True)
+    assert not tmp_output.exists()
+
+
+def test_mask_atmos_plevel_exception():
+    """ if the input file doesnt exist, error """
+    with pytest.raises(FileNotFoundError):
+        mask_atmos_plevel.mask_atmos_plevel_subtool(infile = 'Does not exist',
+                                                    psfile = 'does not exist',
+                                                    outfile = 'will not be created')
```
singhd789: same as the other `warn_no_ps` comment - add it in here (any other places where it might be missing)

fre/app/mask_atmos_plevel/mask_atmos_plevel.py
```diff
@@ -14,55 +23,83 @@ def mask_atmos_plevel_subtool(infile: str, psfile: str, outfile: str) -> None:
...
 
+    fre_logger.info('with xarray, creating a Dataset object')
     ds_out = xr.Dataset()
 
-    # The trigger for atmos masking is a variable attribute "pressure_mask = False".
-    # After the masking is done, change the attribute to True.
-    # Note: these are string types called True and False, not boolean types.
+    fre_logger.debug('The trigger for atmos masking is a variable attribute "pressure_mask = False".')
+    fre_logger.debug('After the masking is done, change the attribute to True.')
+    fre_logger.debug('Note: these are string types called True and False, not boolean types.')
```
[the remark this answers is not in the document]
ilaflott: w/ `emacs` starting from old L48, 
`C - <space>, up up up right right, C - x r t` prompts you to replace a defined rectangle targeting the `# ` lines, input, `fre_logger_debug('`. it's helpful and long-term efficient too because isolates the future carpel tunnel to the left pinky.

fre/app/mask_atmos_plevel/mask_atmos_plevel.py
```diff
@@ -14,55 +23,83 @@ def mask_atmos_plevel_subtool(infile: str, psfile: str, outfile: str) -> None:
...
+    fre_logger.info('with xarray, opening input file %s', infile)
     ds_in = xr.open_dataset(infile)
 
+    fre_logger.info('with xarray, creating a Dataset object')
     ds_out = xr.Dataset()
 
-    # The trigger for atmos masking is a variable attribute "pressure_mask = False".
-    # After the masking is done, change the attribute to True.
-    # Note: these are string types called True and False, not boolean types.
+    fre_logger.debug('The trigger for atmos masking is a variable attribute "pressure_mask = False".')
```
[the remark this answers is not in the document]
ilaflott: i'm gonna cut this out, replace it with shorter `logging` statements, and re-word it and put it in the docstring.

fre/app/mask_atmos_plevel/mask_atmos_plevel.py
ilaflott: `plev` COULD be the vertical coordinate, but not always. the doc string and logic of `pressure_coordinate` clarify this, but poorly. So clarity on this will be added there

fre/app/mask_atmos_plevel/mask_atmos_plevel.py
```diff
@@ -78,34 +115,35 @@ def mask_field_above_surface_pressure(ds: xr.Dataset, var: str, ds_ps: xr.Datase
...
 
-    # broadcast pressure coordinate and surface pressure to
-    # the dimensions of the variable to mask
+    fre_logger.info('broadcasting pressure coordinate and surface pressure to the dimensions of the variable to mask')
     plev_extended, _ = xr.broadcast(plev, ds[var])
     ps_extended, _ = xr.broadcast(ds_ps["ps"], ds[var])
 
-    # read the input file's missing_value
+    # we might need to be more careful about missing_value v _FillValue
+    fre_logger.info('reading the input file\'s missing_value')
```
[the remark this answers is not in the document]
ilaflott: at this point in the logic of the code, we've determined there are variables to mask. the `missing_value` key must exist, as a req. this will be clarified in the doc string.

fre/app/freapp.py
singhd789: Right yeah, I meant more along the lines of could you have just the warning in the code, like do you need the flag too?

ilaflott: the flag changed whether the message gets thrown as a `warning`, and then continues to execute the code, or if it throws the message as an `error`, ceasing the execution.

am i missing the point of the question? 😆

ilaflott: [happy to link you!](https://github.com/NOAA-GFDL/fre-workflows/blob/4b3d0eeb66c28fcdadedbf8e25f9df9e6a0febd4/flow.cylc#L599)

basically, we're targeting every file in the directory in this workflow. we're not using a `diag_manifest.yaml` here (yet), but if we were, we would want to be strict and only open files that WOULD have a `ps` variable available. Without the prior-knowledge provided by something like an external config, we need to skip over chunks of data we can tell we can't adequately process, and not immediately consider that a problem

PR #715, opened 2026-02-02
`fre.cmor`: a CMIP7 example, toward general CMIP7 compatibility 

ilaflott:
## Describe your changes
- adds `cmip7-cmor-tables` as a `git` submodule, right next to `cmip6-cmor-tables`
- adds a CMIP7-flavored example to `fre/tests/test_fre_cmor_cli.py`
- remove `xfail` from tests that now pass (in an intended manner)
- adds a new example (imperfect) user-input configuration for CMIP7
- `cmor_mixer` **is still backwards-compatible for CMIP6 cases** as of this PR. for CMIP7 cases:
  - adds global `GLOBAL_MIP_ERA` to `cmor_mixer` to track which `mip_era` a cmorization call is for
  - adds ability to detect `mip_era`. if CMIP6, function as previous versions did
  - if CMIP7, remove (but track) the brand from variable name in table and move on as normal
  - if CMIP7 and multiple brands possible for one variable, track all brands
  - if CMIP7 and multiple possible brands, check requested dimensions against available dimensions in input file(s)

## Issue ticket number and link (if applicable)
#641, #639

## Checklist before requesting a review

- [ ] I ran my code
- [ ] I tried to make my code readable
- [ ] I tried to comment my code
- [ ] I wrote a new test, if applicable
- [ ] I wrote new instructions/documentation, if applicable
- [ ] I ran pytest and inspected it's output
- [ ] I ran pylint and attempted to implement some of it's feedback
- [ ] No print statements; all user-facing info uses logging module

PR #716, opened 2026-02-02
switch to `numpy>=2`

ilaflott:
numpy v2 compatibility, fixes for NEP 50 type promotion changes. nothing too wild.

### what this does
- bumps `numpy==1.26.4` → `numpy>=2` in `environment.yml` and `meta.yaml`
- fixes numerical precision issues in time averaging routines (`cdoTimeAverager`, `frepytoolsTimeAverager`) caused by numpy 2.0's new type promotion rules
- adds explicit `float64` dtype casting and replaces Python's `sum()` with `numpy.sum()` for consistent results across numpy versions
- adds CI matrix testing for Python 3.12 and 3.13
- drops `Python :: 3.11` classifier from `pyproject.toml`
- some workflow trigger tweaks for `build_conda` and `create_test_conda_env`

### TLDR
copilot agent helped with the heavy lifting on the time averager fixes (#721, #722). the core change is straightforward - just switching the numpy pin - but NEP 50 broke some subtle numerical behavior that needed explicit dtype handling to preserve consistent answers.

### DIFF
--- a/.github/workflows/create_test_conda_env.yml
+++ b/.github/workflows/create_test_conda_env.yml
@@ -15,7 +15,12 @@ concurrency:
 
 jobs:
   condaenvcreate:
+    name: Test Python ${{ matrix.python-version }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.12', '3.13']
     container:
       image: ghcr.io/noaa-gfdl/fre-cli:miniconda24_gcc14_v2
       options: "--privileged --cap-add=sys_admin --cap-add=mknod --device=/dev/fuse --security-opt seccomp=unconfined --security-opt label=disable --security-opt apparmor=unconfined" # needed for podman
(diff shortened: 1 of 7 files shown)


ilaflott:
@copilot prototype a matrix approach for the pipeline scanning python version(s) `3.12`, `3.13`, and `3.14`. I also want to scan for numpy version(s) `=1.26.*` and `>=2.4`.

.github/workflows/create_test_conda_env.yml
```diff
@@ -3,7 +3,7 @@ name: create_test_conda_env
 on:
   pull_request:
     branches:
-      - main
```
singhd789: Why wouldn't we want modifications to run on creation of a PR to main?

ilaflott:
according to #724, this will not clash with #715 functionally

singhd789 (approved):
From what I understood, I think this looks good. Giving approval but not merging as per the conversation in this morning's rewrite meeting about changing dependencies/versions before or after the fre-cli 2026.01 code release

PR #848, opened 2026-04-08
Rename src dir instead of removing

### LINKED ISSUE #844: Rename src/ directory in `fre make checkout --force-checkout` (instead of removing)
**Describe the bug**
When using `fre make checkout --force-checkout`, the src directory and checkout script is fully removed and re-created. However, to keep in line with existing behavior (and to be safer), the directory should just be renames instead of removed.

singhd789:
## Describe your changes
Previously, the `--force-checkout` option removed the checkout script and src_dir that was already generated and re-created them. This did not mimic the current `--force-checkout` behavior in Bronx (which is also a bit more safe). If `--force-checkout` is used, the `src_dir` is renamed (in the same location) to `src.[timestamp]`.

This was only done for bare-metal builds. THe container build doesn't need this as the src code is INSIDE the container. 

- also changed `os` bits to `Path`

Fixes #844  

- [x] I ran my code

### DIFF
--- a/fre/make/create_checkout_script.py
+++ b/fre/make/create_checkout_script.py
@@ -103,8 +103,9 @@ def checkout_create(yamlfile: str, platform: tuple, target: tuple,
     :type njobs: int
     :param execute: If True, run checkout.sh
     :type execute: bool
-    :param force_checkout: If True, for bare-metal: add timestamp to source directory and create new checkout script,
-                           for container: overwrite locally existing checkout script
+    :param force_checkout: If True, for bare-metal build: add timestamp to source directory and create a new checkout script
+                           If True, for container build: overwrite locally existing checkout script before COPY-ing to the 
+                           container image filesystem
     :type force_checkout: bool
 
     :raises ValueError:
--- a/fre/make/create_checkout_script.py
+++ b/fre/make/create_checkout_script.py
@@ -11,8 +11,9 @@ Note, a bare-metal build defaults to a parallel checkout.
 A container build defaults to a non-parallel checkout.
 
 '''
-
-import os
+import shutil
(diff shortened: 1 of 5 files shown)


uramirez8707 (approved):

fre/make/create_checkout_script.py
```diff
@@ -102,7 +103,8 @@ def checkout_create(yamlfile: str, platform: tuple, target: tuple,
     :type njobs: int
     :param execute: If True, run checkout.sh
     :type execute: bool
-    :param force_checkout: If True, overwrite locally existing checkout script and source code
+    :param force_checkout: If True, for bare-metal: add timestamp to source directory and create new checkout script,
+                           for container: overwrite locally existing checkout script
```
mlee03: overwrite locally existing checkout script before COPY-ing to the filesystem in the container image?

fre/make/tests/test_create_checkout.py
```diff
@@ -81,6 +83,30 @@ def test_checkout_execute(monkeypatch):
                 Path(f"{OUT}/fremake_canopy/test/null_model_full/src/coupler").is_dir(),
                 any(Path(f"{OUT}/fremake_canopy/test/null_model_full/src/coupler").iterdir())])
 
+def test_bm_checkout_failure(caplog, monkeypatch):
+    """
+    check for the raised OSError when the checkout script has been run,
```
mlee03: Check for raised OSError when the checkout script is executed twice without the --force-checkout option in the second execution

mlee03 (changes requested):

ceblanton (approved):
Sorry Dana, for letting this chill. It looks great, thanks!

PR #864, opened 2026-04-13
Fix logic to determine non-desired variables in split-vars

### LINKED ISSUE #862: 'fre pp split-netcdf' incorrectly skips some aerosol variables
**Describe the bug**
Certain aerosol variables are being identified as metadata (i.e. not data) variables by `fre pp split-netcdf`, causing them to not be split-out and processed.

These `aerosol_month_cmip` variables are mis-identified:  bldep, pso4_aq_so2_reevap_ls, drybc, emiapoa, emibc, emibvoc, emiisop_biogenic, mmrbc, and wetbc.

**To Reproduce**
Use @uramirez8707 's am5 test case, and check in the `split-netcdf` logs that these particular variables are mis-identified as metadata vars:

[DEBUG:  split_netcdf_script.py:       split_file_xarray] metavars: ['lev_bnds', 'time_bnds', 'lat_bnds', 'lon_bnds', 'ap', 'b', 'ap_bnds', 'b_bnds', 'bldep', 'pso4_aq_so2_reevap_ls', 'drybc', 'emiapoa', 'emibc', 'emibvoc', 'emiisop_biogenic', 'mmrbc', 'wetbc']

**Expected behavior**
These aerosol variables should be split-out normally.

**Additional context**

ceblanton:
The existing logic used both known patterns and a list of variables whose number of dimensions look like metadata. The problem was that the matching logic was applied to the patterns, which was fine, but also the "short var" list, e.g. "a" and "b". This caused any variable with "b" such as "drybc" to be excluded from the output.

Fixes #862  (replace XXX with the issue number and GitHub will autolink the PR to the issue)

- [x] I ran my code

### DIFF
--- a/fre/pp/split_netcdf_script.py
+++ b/fre/pp/split_netcdf_script.py
@@ -15,6 +15,7 @@ import sys
 from itertools import chain
 from os import path
 from pathlib import Path
+from typing import Dict, List, NoReturn, Optional, Union
 
 import xarray as xr
 import yaml
@@ -34,8 +35,15 @@ fre_logger = logging.getLogger(__name__)
 #hurt to double-check.
 METADATA_VAR_PATTERNS = ["_bnds", "_bounds", "_offset", "average_"]
 
-def split_netcdf(inputDir, outputDir, component, history_source, use_subdirs,
-                 yamlfile, split_all_vars=False):
+def split_netcdf(
+    inputDir: str,
+    outputDir: str,
+    component: str,
+    history_source: str,
+    use_subdirs: bool,
+    yamlfile: str,
+    split_all_vars: bool = False,
+) -> NoReturn:
     '''
     Given a directory of netcdf files, splits those netcdf files into separate
     files for each data variable and copies the data variable files of interest
@@ -160,7 +168,9 @@ def split_netcdf(inputDir, outputDir, component, history_source, use_subdirs,
     fre_logger.info(f"split-netcdf-wrapper call complete, having split {files_split} files")
     sys.exit(0) #check this
 
-def split_file_xarray(infile, outfiledir, var_list='all'):
+def split_file_xarray(
+    infile: str, outfiledir: str, var_list: Union[str, List[str]] = "all"
+) -> None:
     '''
     Given a netcdf infile containing one or more data variables,
     writes out a separate file for each data variable in the file, including the
@@ -205,7 +215,7 @@ def split_file_xarray(infile, outfiledir, var_list='all'):
     fre_logger.info(f"To exclude: var patterns matching '{METADATA_VAR_PATTERNS}'")
     fre_logger.info(f"To exclude: 1 or 2-d vars: '{metadata_vars_to_exclude_by_name}'")
     #both combined gets you a decent list of non-diagnostic variables
-    def is_metadata_var(var_to_check):
+    def is_metadata_var(var_to_check: str) -> bool:
         """
         Checks a variable name and determines whether it is a metadata variable
         and therefore should not be one of the split-by-variable output files, by
@@ -270,7 +280,7 @@ def split_file_xarray(infile, outfiledir, var_list='all'):
             data2.to_netcdf(var_out, encoding = var_encode)
             fre_logger.debug(f"Wrote '{var_out}'")
 
-def get_max_ndims(dataset):
+def get_max_ndims(dataset: xr.Dataset) -> int:
     '''
     Gets the maximum number of dimensions of a single var in an xarray
     Dataset object. Excludes coord vars, which should be single-dim anyway.
@@ -284,7 +294,7 @@ def get_max_ndims(dataset):
     ndims = [len(dataset[v].shape) for v in allvars]
     return max(ndims)
 
-def set_coord_encoding(dset, vcoords):
+def set_coord_encoding(dset: xr.Dataset, vcoords: List[str]) -> Dict[str, Dict[str, Union[None, str, type]]]:
     '''
     Gets the encoding settings needed for xarray to write out the coordinates
     as expected
@@ -318,7 +328,7 @@ def set_coord_encoding(dset, vcoords):
             encode_dict[vc]['units'] = dset[vc].encoding['units']
     return encode_dict
 
-def set_var_encoding(dset, varnames):
+def set_var_encoding(dset: xr.Dataset, varnames: List[str]) -> Dict[str, Dict[str, Union[None, str, type]]]:
     '''
     Gets the encoding settings needed for xarray to write out the variables
     as expected
@@ -345,7 +355,7 @@ def set_var_encoding(dset, varnames):
             encode_dict[v]['units'] = dset[v].encoding['units']
     return encode_dict
 
-def fre_outfile_name(infile, varname):
+def fre_outfile_name(infile: str, varname: str) -> str:
     '''
     Builds split  var filenames the way that fre expects them
     (and in a way that should work for any .nc file)
--- a/fre/pp/split_netcdf_script.py
+++ b/fre/pp/split_netcdf_script.py
@@ -32,7 +32,7 @@ fre_logger = logging.getLogger(__name__)
 #*_average: calculated averages for a variable.
 #These vars may also be covered by the var_shortvars query, but it doesn't
 #hurt to double-check.
-VAR_PATTERNS = ["_bnds", "_bounds", "_offset", "average_"]
+VAR_EXCLUDE_PATTERNS = ["_bnds", "_bounds", "_offset", "average_"]
 
 def split_netcdf(inputDir, outputDir, component, history_source, use_subdirs,
                  yamlfile, split_all_vars=False):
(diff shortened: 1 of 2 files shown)


fre/pp/split_netcdf_script.py
```diff
@@ -198,21 +198,30 @@ def split_file_xarray(infile, outfiledir, var_list='all'):
     #If they were, I could get away with the following:
     #var_zerovars = [v for v in datavars if not len(dataset[v].coords) > 0])
     #instead of this:
-    var_shortvars = [v for v in allvars if (len(dataset[v].shape) < varsize) and v not in dataset._coord_names]
+    shortvars_to_exclude = [v for v in allvars if (len(dataset[v].shape) < varsize) and v not in dataset._coord_names]
+    fre_logger.debug(f"Variables to be excluded (due to small number of dimensions): '{shortvars_to_exclude}'")
```
mlee03: short-named variables to exclude in search?

fre/pp/split_netcdf_script.py
```diff
@@ -198,21 +198,30 @@ def split_file_xarray(infile, outfiledir, var_list='all'):
...
-
-            xstr: string to search for matches
-            var_exclude: list of patterns defined in VAR_EXCLUDE'''
-        allmatch = [re.search(el, xstr)for el in var_exclude]
-        #If there's at least one match in the var_exclude list (average_bnds is OK)
-        return len(list(set(allmatch))) > 1
+        """
+        Checks a string for matches in patterns or exact names.
+
+        xstr: string to search for matches
```
mlee03: type hinting

fre/pp/split_netcdf_script.py
```diff
@@ -198,21 +198,30 @@ def split_file_xarray(infile, outfiledir, var_list='all'):
...
     def matchlist(xstr):
-        ''' checks a string for matches in a list of patterns
-
-            xstr: string to search for matches
-            var_exclude: list of patterns defined in VAR_EXCLUDE'''
-        allmatch = [re.search(el, xstr)for el in var_exclude]
-        #If there's at least one match in the var_exclude list (average_bnds is OK)
-        return len(list(set(allmatch))) > 1
+        """
+        Checks a string for matches in patterns or exact names.
```
mlee03: it's not clear what string is being matched to what..

fre/pp/split_netcdf_script.py
```diff
@@ -198,21 +198,30 @@ def split_file_xarray(infile, outfiledir, var_list='all'):
...
+
+        xstr: string to search for matches
+        VAR_EXCLUDE_PATTERNS: list of patterns defined to exclude from output
```
mlee03: maybe this documentation can go at the top of the module?

mlee03: list of patterns to exclude when regex matching patterns in a given string

fre/pp/split_netcdf_script.py
```diff
@@ -198,21 +198,30 @@ def split_file_xarray(infile, outfiledir, var_list='all'):
...
+        shortvars_to_exclude: list of variables to exclude from output
```
mlee03: it's unclear what output is

mlee03 (changes requested):

fre/pp/split_netcdf_script.py
```diff
@@ -198,21 +198,30 @@ def split_file_xarray(infile, outfiledir, var_list='all'):
...
+    fre_logger.debug(f"Variables to be excluded (due to small number of dimensions): '{shortvars_to_exclude}'")
     #having a variable listed as both a metadata var and a coordinate var seems to
     #lead to the weird adding a _FillValue behavior
-    fre_logger.info(f"var patterns: {VAR_PATTERNS}")
-    fre_logger.info(f"1 or 2-d vars: {var_shortvars}")
+    fre_logger.info(f"To exclude: var patterns matching '{VAR_EXCLUDE_PATTERNS}'")
+    fre_logger.info(f"To exclude: 1 or 2-d vars: '{shortvars_to_exclude}'")
     #both combined gets you a decent list of non-diagnostic variables
-    var_exclude = list(set(VAR_PATTERNS + [str(el) for el in var_shortvars] ))
     def matchlist(xstr):
```
mlee03: Maybe the function name can be a bit more descriptive?

fre/pp/split_netcdf_script.py
cwhitlock-NOAA: The *names* of the variables aren't what are short - the variables are "short" because they reference a small number of dimensions, which makes it likely that they are getting used as metadata (i.e. a coordinate variable, a value in an equation, or a bounds variable)

fre/pp/split_netcdf_script.py
cwhitlock-NOAA: Suggested comment to add: "This module is intended to be used only internally within this function, and returns a list of true/false values for the list comprehensions around lines 225/226. Values are TRUE if the criteria for a metadata-like variable are met (the two checked cases) and FALSE if they are not (the fall-through case)"

fre/pp/split_netcdf_script.py
```diff
@@ -198,21 +198,30 @@ def split_file_xarray(infile, outfiledir, var_list='all'):
...
+        VAR_EXCLUDE_PATTERNS: list of patterns defined to exclude from output
+        shortvars_to_exclude: list of variables to exclude from output
+        """
+        # Check substring patterns from VAR_EXCLUDE_PATTERNS
+        for pattern in VAR_EXCLUDE_PATTERNS:
+            if re.search(pattern, xstr):
+                return True
+        # Check exact matches from shortvars_to_exclude
+        for name in shortvars_to_exclude:
+            if xstr == name:
```
cwhitlock-NOAA: So to make sure I'm understanding the problem...

Formula terms like "a" and "b" that have a single dimension were previously getting classified as shortvars_to_exclude and then added to the VAR_EXCLUDE_PATTERNS, which lead to all variables with an a or a b in the string to be classed as metadata-like variables. The solution was to say that we wanted different levels of pattern-matching for the VAR_EXCLUDE_PATTERNS and the shortvars_to_exclude - VAR_EXCLUDE_PATTERNS should be a pattern-matching and shortvars_to_exclude is an exact match. 

This looks perfectly reasonable given my understanding of what's going on - I might want to add a bit more documentation on the var classification, but that's a separate PR.

cwhitlock-NOAA (approved):
So to make sure I'm understanding the problem...

Formula terms like "a" and "b" that have a single dimension were previously getting classified as shortvars_to_exclude and then added to the VAR_EXCLUDE_PATTERNS, which lead to all variables with an a or a b in the string to be classed as metadata-like variables. The solution was to say that we wanted different levels of pattern-matching for the VAR_EXCLUDE_PATTERNS and the shortvars_to_exclude - VAR_EXCLUDE_PATTERNS should be a pattern-matching and shortvars_to_exclude is an exact match.

This looks perfectly reasonable given my understanding of what's going on - I might want to add a bit more documentation on the var classification, but that's a separate PR.

fre/pp/split_netcdf_script.py
ceblanton: How about `metadata_vars` or `non_diagnostic_vars`? I agree "short" is a bit unclear.

fre/pp/split_netcdf_script.py
ceblanton: Exactly. Thanks!

fre/pp/split_netcdf_script.py
ceblanton: `metadata_vars_to_exclude_by_name`

I think it's more clear..

fre/pp/split_netcdf_script.py
ceblanton: Thanks, good point.

`is_metadata_var` is a bit better

fre/pp/split_netcdf_script.py
ceblanton: Thanks both! added

fre/pp/split_netcdf_script.py
ceblanton: done!

fre/pp/split_netcdf_script.py
```diff
@@ -198,23 +208,39 @@ def split_file_xarray(infile, outfiledir, var_list='all'):
...
+        Checks a variable name and determines whether it is a metadata variable
+        and therefore should not be one of the split-by-variable output files, by
+        comparing the variable name to exact matches or pattern matches of known metadata variables.
+
+        This nested method is intended to be used only internally within this method (split_file_xarray),
+        and returns a list of true/false values for the list comprehensions around lines 225/226.
+        Values are TRUE if the criteria for a metadata-like variable are met (the two checked cases)
+        and FALSE if they are not (the fall-through case)
+
+        xstr: string to search for matches
```
singhd789: Looks like you need to update `xstr` to `var_to_check` and add 
```
:param var_to_check: string to search for matches
:type var_to_check: str
```

fre/pp/split_netcdf_script.py
ceblanton: `METADATA_VAR_PATTERNS` is used only in `split_file_xarray` so I moved it there (with its comments)

fre/pp/split_netcdf_script.py
ceblanton: Thanks Mikyung. These are great catches. I _think_ I'm improving them the wording here..

METADATA_VAR_PATTERNS: list of regex patterns defined that match variables that should be excluded from the split-out files.
metadata_vars_to_exclude_by_name: list of variable names that should be excluded from the split-out files

fre/pp/split_netcdf_script.py
ceblanton: Thank you, agreed

singhd789 (commented):
Do any tests need to be adjusted or updated here?

ceblanton:
You're absolutely right, Dana. I'll add a test to ensure that variables containing the letters "a" or "b" are processed.

ceblanton:
thanks @ilaflott for the fix in #883 which also fixed this PR pipeline

fre/pp/tests/test_split_netcdf.py
```diff
@@ -219,6 +220,72 @@ def test_split_file_metadata(workdir,newdir, origdir):
...
+    with tempfile.TemporaryDirectory() as tmpdir:
+        # Call split_file_xarray with "all" to get all data vars
+        split_file_xarray.__wrapped__ = split_file_xarray  # Access the original function
+        
+        # We need to test the internal logic. Since split_file_xarray writes files,
+        # let's test the filtering logic by inspecting the datavars.
+        # Actually, let's modify the function to return the datavars for testing.
+        
+        # For now, let's call it and check the output files
+        output_file = osp.join(tmpdir, 'test.nc')
```
singhd789: Is `output_file` used anywhere?

singhd789: (besides this clarification, from what I understand, I think this PR looks good)

ceblanton: Thanks very much for the catch, Dana. you are correct! It is a dangling variable never used. I'll fix

singhd789 (approved):

PR #872, opened 2026-04-17
 Set `DO_ANALYSIS=False` in rose-suite if no analysis at all

### LINKED ISSUE #847: Set default to `postprocess_on` and `analysis_on`
When defining the PP components, the `postprocess_on` key is always mandatory. 

This is is going to lead to a lot of repetition (and longer yamls).

I think the `postprocess_on` key should be option and the default should be True. The only case where you would need that key is when `postprocess_on: False`

**Describe alternatives you've considered**
We can leave it as is. This is not a problem, just an inconvenience to the users.

**Additional context**
A similar change should happen with the `analysis_on` key

singhd789:
## Describe your changes
Previously, the function exits if there is no analysis section defined in the yaml at all but I forgot that the workflow will still look for the DO_ANALYSIS key set in the rose_suite.conf. 

- added line to set DO_ANALYSIS if no analysis section is defined 

Fixes #847 

- [x] I ran my code

### DIFF
--- a/fre/pp/configure_script_yaml.py
+++ b/fre/pp/configure_script_yaml.py
@@ -188,15 +188,18 @@ def set_rose_suite(yamlfile: dict, rose_suite: metomi.rose.config.ConfigNode) ->
                         value = 'False' )
 
     # Set DO_ANALYSIS switch
-    # analysis_on is optional key for each component in the analysis yaml and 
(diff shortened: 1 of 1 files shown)


ceblanton (approved):
Definitely. Thank you for catching this!

PR #873, opened 2026-04-17
Directories should be added to rose suite

singhd789:
## Describe your changes
....I'm so sorry....the directories should be added before the function potentially exits (due to a non-existent analysis section). This moves the directory information being added to the rose-suite first. 😬  (this would probably be caught earlier when we add the ability to specify a fre-cli fork to test with fre-workflows in the fre-workflow cloud runner test)

- [ ] I ran my code


ceblanton (approved):
Yes, agreed. With the short-circuit `return` you added, any setting after that `return` must go earlier or else it won't be run.

Also: we are (well, you are :) actively working the replacement of this rose-suite Cylc definition configuration in favor of Cylc directly reading the serialized yaml, so this method won't be used long-term anyway.

Thanks!

PR #880, opened 2026-04-24
Fix cont tag

### LINKED ISSUE #879: Container tag not being set correctly
**Describe the bug**
The container tag seems to not be getting correctly set. Look's like the git tag isn't being grabbed correctly and the default 'latest' tag is always being set.

**To Reproduce**

**Expected behavior**

**Additional context**

Ciheim:
## Describe your changes
Adjusts tag setting logic for container auto-build

Fixes #879  (replace XXX with the issue number and GitHub will autolink the PR to the issue)

- [x] I ran my code

### DIFF
--- a/.github/workflows/update_container_image.yaml
+++ b/.github/workflows/update_container_image.yaml
@@ -37,7 +37,7 @@ jobs:
         id: set_tag
         run: |
           TAG=$(git describe --tags --abbrev=0)
-          echo "docker_tag=$TAG" >> $GITHUB_OUTPUT
+          echo "container tag=$TAG" >> $GITHUB_OUTPUT
 
       - name: Log in to the Container registry
         uses: docker/login-action@v3
--- a/.github/workflows/update_container_image.yaml
+++ b/.github/workflows/update_container_image.yaml
@@ -30,18 +30,14 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Determine image tag
+      - name: Fetch all tags
+        run: git fetch --tags
+
+      - name: Determine latest tag
         id: set_tag
         run: |
-          # Use the pushed tag if this is a tag ref; otherwise fall back.
-          if [[ "$GITHUB_REF" == refs/tags/* ]]; then
-            TAG="${GITHUB_REF#refs/tags/}"
-          elif [[ "${GITHUB_REF_NAME}" == "main" ]]; then
-            TAG="latest"
-          else
-            TAG="${GITHUB_SHA::7}"
-          fi
-          echo "docker_tag=$TAG" >> "$GITHUB_OUTPUT"
+          TAG=$(git describe --tags --abbrev=0)
+          echo "docker_tag=$TAG" >> $GITHUB_OUTPUT
 
       - name: Log in to the Container registry
         uses: docker/login-action@v3
--- a/.github/workflows/update_container_image.yaml
+++ b/.github/workflows/update_container_image.yaml
@@ -36,8 +36,7 @@ jobs:
       - name: Determine latest tag
         id: set_tag
         run: |
-          TAG=$(git describe --tags --abbrev=0)
-          echo "container tag=$TAG" >> $GITHUB_OUTPUT
+          echo "docker_tag=$(git describe --tags --abbrev=0)" >> "$GITHUB_OUTPUT"
 
       - name: Log in to the Container registry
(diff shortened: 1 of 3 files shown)


.github/workflows/update_container_image.yaml
```diff
@@ -30,18 +30,14 @@ jobs:
...
-          if [[ "$GITHUB_REF" == refs/tags/* ]]; then
-            TAG="${GITHUB_REF#refs/tags/}"
-          elif [[ "${GITHUB_REF_NAME}" == "main" ]]; then
-            TAG="latest"
-          else
-            TAG="${GITHUB_SHA::7}"
-          fi
-          echo "docker_tag=$TAG" >> "$GITHUB_OUTPUT"
+          TAG=$(git describe --tags --abbrev=0)
+          echo "container tag=$TAG" >> $GITHUB_OUTPUT
```
singhd789: Just clarifying - is `GITHUB_OUTPUT` used anywhere?

Ciheim: Yes, it's how we retrieve the tag from 'steps.set_tag.outputs.docker_tag'. $GITHUB_OUTPUT is important for passing info between steps.

.github/workflows/update_container_image.yaml
singhd789: ok sorry, one more clarifying question - is repository dispatch needed for `types: [publish-conda-succeeded]`? I think repository dispatch is usually used for pipelines in other repos right? But since the other pipeline this one is dependent on is in this repo, can't `workflow_run` be used? (I think I've seen that before). However, not sure if that makes it so it relies on the success of the other pipeline...

Ciheim: No, repo dispatching is not needed! Workflow_run works fine here and I've made this switch. Thanks!

singhd789, line 1: One more thing before I head home for the day - is there any way to check the container built with the correct tag with the `dry-run` option in the manual trigger? If not, maybe that's a useful thing to add right after the last job in the pipeline. (if possible I guess, but I also don't mean to prolong this getting in, it's looking good!) Looks like the latest pipeline was successful! 🎊

.github/workflows/update_container_image.yaml
```diff
@@ -10,8 +10,9 @@ on:
         type: choice
         options: ["true", "false"]
```
singhd789: Reminder: Just remove the `false` option if this is just for testing

Ciheim: Done, thanks!

.github/workflows/update_container_image.yaml
Ciheim: Yes, in the 'show image tag' job there is an echo to display the tag being used

.github/workflows/update_container_image.yaml
```diff
@@ -30,32 +23,35 @@ jobs:
...
-          fi
-          echo "docker_tag=$TAG" >> "$GITHUB_OUTPUT"
+          echo "docker_tag=$(git describe --tags --abbrev=0)" >> "$GITHUB_OUTPUT"
+
+      - name: Show image tag
+        run: echo "Docker tag is ${{ steps.set_tag.outputs.docker_tag }}"
+
+      - name: Update fre-cli version in cylc-flow-tools.yaml
+        run: |
+          sed -i 's/^[[:space:]]*-[[:space:]]*noaa-gfdl::fre-cli==.*/  - noaa-gfdl::fre-cli==${{ steps.set_tag.outputs.docker_tag }}/' ./container-files/cylc-flow-tools.yaml
```
ceblanton, line 39: cylc-flow-tools.yaml still contains this old tag

```
dependencies:
  - noaa-gfdl::fre-cli==2025.04
```

Why not commit the version change update to `cylc-flow-tools.yaml`?

ceblanton (changes requested):
Could you link the issue that this PR addresses, and check the PR template boxes?

It's a CI-only update but is good practice!

.github/workflows/update_container_image.yaml
singhd789: Going off of this - the hsm package could also be tagged to 1.4.0 for that hsm update (though the container that is built won't work due to the lack of fregrid - unrelated to hsm)

ceblanton (approved):

PR #909, opened 2026-06-11
908.fix-split-netcdf-race-bug

ceblanton:
## Describe your changes

The CI job "Test Python 3.14" was failing with `FAILED fre/pp/tests/test_split_netcdf.py::test_makedirs_subdir_preexists - SystemExit: 0`.

**Root cause:** The `split_netcdf` function in `fre/pp/split_netcdf_script.py` contained a `sys.exit(0)` call (marked `#check this`) at the end of the function body. When the test `test_makedirs_subdir_preexists` called this function directly (not via CLI), `sys.exit(0)` raised `SystemExit: 0`, which pytest treated as a test failure.

**Changes made (`fre/pp/split_netcdf_script.py`):**
- Removed `sys.exit(0)` so `split_netcdf` returns normally instead of exiting the process
- Updated return type annotation from `NoReturn` to `None` to accurately reflect the function's behaviour
- Removed now-unused `sys` import and `NoReturn` from the `typing` imports

- [x] I ran my code


ceblanton:
@singhd789 this is definitely what Niki ran into yesterday. I didn't really test it hard but we know it's right..

singhd789 (commented):
Besides the submodule commit bit in the failing tests, this looks good! Thanks Chris.

ceblanton:
Oh my goodness. Thank you for the catch Dana. I foolishly used "git commit -a". I will remove

ceblanton:
Fixed! thank you again for the catch there

singhd789:
Ooooh man, hate to say it, maybe update/have a little test for the line change for the codecov failure that copilot can write up?

fre/pp/split_netcdf_script.py
```diff
@@ -129,8 +129,7 @@ def split_netcdf(
                 fre_logger.info(f"No input files found; skipping subdir {sd}")
             else:
                 output_subdir = os.path.join(os.path.abspath(outputDir), sd)
-                if not os.path.isdir(output_subdir):
-                    os.mkdir(output_subdir)
+                os.mkdir(output_subdir, exist_ok=True)
```
ilaflott: this doesn't work but the pipeline passes because it's uncovered
```
>>> os.mkdir('./foo/',exist_ok=True)
Traceback (most recent call last):
  File "<python-input-3>", line 1, in <module>
    os.mkdir('./foo/',exist_ok=True)
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: mkdir() got an unexpected keyword argument 'exist_ok'
```

ilaflott (changes requested):

fre/pp/split_netcdf_script.py
ilaflott: `os.makedirs('./foo',exist_ok=True)` is what you want

i was a little worried it could clobber output, but it doesn't- good!

```
 ~ $] mkdir foo
 ~ $] touch foo/bar
 ~ $] python
Python 3.14.5 | packaged by conda-forge | (main, May 20 2026, 00:15:36) [GCC 14.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.makedirs('./foo',exist_ok=True)
>>>
 ~ $] ls foo
bar
```

singhd789: Thanks for this catch. The codecov failing did make me rethink some things there

singhd789: I think it would either be `os.makedirs('foo', exist_ok)` or with pathlib `Path(foo).mkdir(parents=True, exist_ok=True)`

ceblanton: Thanks both. What a comedy of errors I had this afternoon. zero of 2 for one line changes

ceblanton: I'm blushing from embarrassment. It's fixed now with a test case

ceblanton:
> Ooooh man, hate to say it, maybe update/have a little test for the line change for the codecov failure that copilot can write up?

You know, I would not have done this without your suggestion, but it was so easy, and clearly the right thing to do.

fre/pp/tests/test_split_netcdf.py
```diff
@@ -285,6 +285,40 @@ def test_variable_filtering_bug_fix():
...
+    split_netcdf_script.split_netcdf(
+        inputDir=str(tmp_path / "input"),
+        outputDir=str(output_dir),
+        component="atmos",
+        history_source=history_source,
+        use_subdirs=True,
+        yamlfile=None,
+        split_all_vars=True,
+    )
+
```
singhd789, line 320: If the `os.makedirs` line spits out a FileExists error, I think you'll want to wrap the split-netcdf script call in something like this for the check:
```
  with pytest.raises( FileExistsError ) as execinfo:
      split_netcdf_script....
  assert execinfo.type is FileExistsError
```

singhd789:
> > Ooooh man, hate to say it, maybe update/have a little test for the line change for the codecov failure that copilot can write up?
> 
> You know, I would not have done this without your suggestion, but it was so easy, and clearly the right thing to do.

It was Ian's catch as well!  :laughing:  He's got an eye for it

fre/pp/tests/test_split_netcdf.py
singhd789: Nevermind! I misread the description

singhd789:
Note: for some reason the create pytest took`sys.exit(0)` as a failure, so it was commented out (probably can be removed though)

singhd789 (approved):

ilaflott (approved):
looks good. test too!

PR #912, opened 2026-06-11
update for node24 migration

ilaflott:
## Describe your changes
actions will now need to use node js 24 as node js 20 is being phased out (EOL)

this PR updates actions to use node js 24 github runners now. minimal friction expected, but better safe than sorry. 

see the warnings attached to any workflow without the edits in this PR


ilaflott:
note this only affects the github repository, and not any built/published packages

ceblanton (approved):

PR #922, opened 2026-06-25
Update publish-conda to use the new tag for the package version

### LINKED ISSUE #921: new version releases point to the previous tag
**Describe the bug**
When a tag is pushed, publish_conda and update_tag both fire at the same time. publish_conda immediately builds the conda package from the tagged commit, but at that point `__init__.py` still has the previous tag hardcoded as the fallback version string — update_tag hasn't had a chance to open its PR, let alone get it merged. So the published conda package installs an `__init__.py` with the wrong fallback, and users who run `fre --version` (where GIT_DESCRIBE_TAG is not set) see the old version instead of the new one.

**To Reproduce**
Trigger a new fre-cli release by pushing a new tag. The resulting conda package will be associated with the previous tag (`fre --version`)

**Expected behavior**
After triggering a new fre-cli release with a new tag, the published conda package should report the new tag in `fre --version`.

**Additional context**

ceblanton:
## Describe your changes
Run the publish-conda action on pushes to main or after the update-tag action has run.

We detect the latter case by a closed pull request on a main push AND the if-condition that checks that the source branch is the one triggered by the update-tag reference changes.

The "Set version for tagged release" is needed to inform conda-build of the tag. Because there are intermediate commits (the updated version refs) the commit no longer is tagged. So we can retrieve the package version from __init__.py

Fixes #921

- [ ] I ran my code

### DIFF
--- a/.github/workflows/publish_conda.yml
+++ b/.github/workflows/publish_conda.yml
@@ -3,15 +3,30 @@ name: publish_conda
 env:
   FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 
+# Run this action on pushes to main or after the update-tag action has run.
+# We detect the latter case by a closed pull request on a main push
+#   AND the if-condition that checks that the source branch is the
+#   one triggered by the update-tag reference changes.
+# The "Set version for tagged release" is needed to inform conda-build
+#   of the tag. Because there are intermediate commits (the updated version refs)
+#   the commit no longer is tagged. So we can retrieve the package version
+#   from __init__.py.
 on:
   push:
     branches:
       - main
-    tags:
-      - '*'
+  pull_request:
+    types: [closed]
+    branches:
+      - main
 
 jobs:
   condapublish:
+    if: |
+      github.event_name == 'push' ||
+      (github.event_name == 'pull_request' &&
+       github.event.pull_request.merged == true &&
+       startsWith(github.event.pull_request.head.ref, 'update-version-post-release'))
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/noaa-gfdl/fre-cli:miniconda24_gcc14_v2
@@ -20,6 +35,17 @@ jobs:
         uses: actions/checkout@v4
         with:
           submodules: 'recursive'
+
+      - name: Set version for tagged release
+        if: github.event_name == 'pull_request'
+        run: |
+          VERSION=$(python3 -c "
+          import re
+          content = open('fre/__init__.py').read()
+          m = re.search(r'GIT_DESCRIBE_TAG\",\s*\"([^\"]+)\"', content)
+          print(m.group(1))
+          ")
+          echo "GIT_DESCRIBE_TAG=$VERSION" >> $GITHUB_ENV
 
       - name: Configure Conda
         run: |
--- a/.github/workflows/publish_conda.yml
+++ b/.github/workflows/publish_conda.yml
@@ -36,6 +36,12 @@ jobs:
         with:
           submodules: 'recursive'
 
+      # For tagged releases, determine the tag from the __init__.py.
+      # Then make it available for the conda-build steps by setting the env var
+      # $GIT_DESCRIBE_TAG to the determined tag.
+      # $GITHUB_ENV is a special file path provided by the Actions runner.
+      # When you write KEY=VALUE lines into it, those variables become available
+      # as environment variables for all subsequent steps in the same job.
       - name: Set version for tagged release
         if: github.event_name == 'pull_request'
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_conda.yml
```diff
@@ -3,15 +3,30 @@ name: publish_conda
...
 on:
   push:
     branches:
       - main
-    tags:
-      - '*'
+  pull_request:
+    types: [closed]
+    branches:
+      - main
```
singhd789, line 21: Ok so with this, what happens if a PR is abandoned or becomes old and someone closes it though? We wouldn't want a whole container build for that right?

.github/workflows/publish_conda.yml
```diff
@@ -3,15 +3,30 @@ name: publish_conda
...
 
 jobs:
   condapublish:
+    if: |
+      github.event_name == 'push' ||
+      (github.event_name == 'pull_request' &&
+       github.event.pull_request.merged == true &&
+       startsWith(github.event.pull_request.head.ref, 'update-version-post-release'))
```
singhd789: Should the spacing be slightly off here?

.github/workflows/publish_conda.yml
```diff
@@ -21,6 +36,17 @@ jobs:
...
+        if: github.event_name == 'pull_request'
+        run: |
+          VERSION=$(python3 -c "
+          import re
+          content = open('fre/__init__.py').read()
+          m = re.search(r'GIT_DESCRIBE_TAG\",\s*\"([^\"]+)\"', content)
+          print(m.group(1))
+          ")
+          echo "GIT_DESCRIBE_TAG=$VERSION" >> $GITHUB_ENV
+
```
singhd789, line 55: Would this effect the fre-cli container that is tagged with "develop"? How is that one built again? I think @ilaflott explained it before but I may need a reminder

.github/workflows/publish_conda.yml
ceblanton: Thanks Dana for the eyes. I had the same reaction you did. But this action will only run if the pull request is closed AND if the main branch is targeted.

So a simply closed PR will not be triggered as it's not targeting the main branch.

.github/workflows/publish_conda.yml
ceblanton: It does look a little off crammed together like that. You'd vote for the continuation lines to be indented, like this?

    if: |
      github.event_name == 'push' ||
          (github.event_name == 'pull_request' &&
          github.event.pull_request.merged == true &&
          startsWith(github.event.pull_request.head.ref, 'update-version-post-release'))

.github/workflows/publish_conda.yml
ceblanton: Good question.. we definitely want to preserve the "develop" deployment.

This "Set version for tagged release" step will only run on pull requests, not pushes to main.

The only impact of that step is echo'ing the git tag to the $GITHUB_ENV location whereever that is.

But how is $GITHUB_ENV used? I agree, it's a little opaque.

.github/workflows/publish_conda.yml
singhd789: Ok follow up, wouldn't the `main` branch mainly be targeted for a pull request?

.github/workflows/publish_conda.yml
singhd789: Just being a bit nit picky, but personally, I would prefer that for just a bit more readability.

.github/workflows/publish_conda.yml
ceblanton: Alright, I see now. Here's the explanation:

For "develop" package releases on pushes to main, the "Set version for tagged release" step will not run. When the "Build fre-cli conda package" step runs, `conda build .` defaults to package version "develop" because it is (spookily) git aware when run in this context.

For tagged package releases, the "Set version for tagged release" step _will_ run. This $GITHUB_ENV variable is a special file path provided by github actions. When you write KEY=VALUE to it, those variables become environment variables for all subsequent steps.

So for tagged releases, when "Build fre-cli conda package" runs, "conda build ." sees the $GIT_DESCRIBE_TAG variable and uses it as if it were a proper tag. (That commit is one past the tag, because the tag trigger the __init__.py change..)

.github/workflows/publish_conda.yml
ceblanton: You're asking, why wouldn't a "pull request" only trigger be sufficient?

You answered it! Because most, but not all, pull requests are to main.

.github/workflows/publish_conda.yml
ceblanton: I'll adjust

.github/workflows/publish_conda.yml
ceblanton: I added a inline comment since this was _extremely_ confusing I couldn't agree more.

Thanks so much for raising this.

singhd789 (approved):

Issue #955, opened 2026-07-23
Update `fre list ` subtools to account for incoming yaml changes and update documentation


singhd789:
**Is your feature request related to a problem? Please describe.**
1. update documentation for fre-cli chatbot
2. update subtools to account for format change from integrating uw tools
3. strip out any references to am5 in yaml examples

PR #959, opened 2026-07-28
Update doc strings for list subtools

singhd789:
## Describe your changes
Update documentation for fre list subtools to fit the format for the fre-cli chatbot

Starts #955

- [ ] I ran my code

### DIFF
--- a/fre/list_/list_experiments_script.py
+++ b/fre/list_/list_experiments_script.py
@@ -1,6 +1,11 @@
 """
-Script combines the model yaml with experiment, platform, and target 
-information to list experiment information.
+List_experiments_script provides a method to list experiments
+found in the `model.yaml`. The yaml is resolved in case experiment
+names use fre_properties.
+
+
+The experiments should mainly be listed and defined in the
+`model.yaml`, therefore it is the only yaml needed to parse.
 """
 
 import logging
@@ -18,9 +23,10 @@ fre_logger = logging.getLogger(__name__)
 
 def list_experiments_subtool(yamlfile: str):
     """
-    List the post-processing experiments available
+    List_experiments_subtool loads and parses the model yaml configuration
+    to list the experiments defined.
 
-    :param yamlfile: path to yaml configuration file
+    :param yamlfile: is the path to model.yaml configuration file
     :type yamlfile: str
     """
     exp = None
--- a/fre/list_/list_platforms_script.py
+++ b/fre/list_/list_platforms_script.py
@@ -1,9 +1,9 @@
 """
 List_platforms_script provides a method to list platforms defined in
 the platforms.yaml. With this subtool, the `model.yaml` is passed and
-combined with the `compile.yaml` and `platforms.yaml` in case the
-`platforms.yaml` uses `fre_properties` references defined in the two
-previous YAML configurations.
+combined with the `compile.yaml` and `platforms.yaml`. This merge is
+done in case the `platforms.yaml` uses `fre_properties` references
+defined in the two aformentioned YAML configurations.
 """
 
 from pathlib import Path
--- a/fre/list_/list_platforms_script.py
+++ b/fre/list_/list_platforms_script.py
@@ -1,6 +1,9 @@
 """
-Script combines the model yaml with the compile and platforms yaml.
-Merged yaml dictionary is parsed to list experiment information.
+List_platforms_script provides a method to list platforms defined in
+the platforms.yaml. With this subtool, the `model.yaml` is passed and
+combined with the `compile.yaml` and `platforms.yaml` in case the
+`platforms.yaml` uses `fre_properties` references defined in the two
+previous YAML configurations.
 """
 
 from pathlib import Path
@@ -12,9 +15,12 @@ fre_logger = logging.getLogger(__name__)
 
 def list_platforms_subtool(yamlfile: str):
     """
-    List the platforms available
+    List_platforms_subtool uses the "fre yamltools combine-yamls"
+    subtool to combine the model, compile, and platform YAML
+    configurations in order to parse a fully resolved YAML configuration
+    to list the platforms available/defined in the `platforms.yaml`.
 
-    :param yamlfile: path to the yaml configuration file
+    :param yamlfile: is the path to the model.yaml configuration file
     :type yamlfile: str
     """
     # set logger level to INFO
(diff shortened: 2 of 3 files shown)


fre/list_/list_platforms_script.py
```diff
@@ -1,6 +1,9 @@
 """
-Script combines the model yaml with the compile and platforms yaml.
-Merged yaml dictionary is parsed to list experiment information.
+List_platforms_script provides a method to list platforms defined in
+the platforms.yaml. With this subtool, the `model.yaml` is passed and
+combined with the `compile.yaml` and `platforms.yaml` in case the
+`platforms.yaml` uses `fre_properties` references defined in the two
+previous YAML configurations.
```
Ciheim: Are the 'two previous YAML configurations' the model and compile yaml? I feel like the right word would be 'aforementioned' instead of previous.

fre/list_/list_pp_components_script.py
```diff
@@ -1,6 +1,12 @@
...
-Script combined model yaml with the settings and post-processing yamls.
-Merged yaml dictionary is parsed to list components that will be post-processed
+List_ppcomps_subtool provides a method to list components to be post-processed,
+defined in the post-processing YAML configurations. The "fre yamltools combine-yamls"
+subtool is used to help resolve any aliases defined in the configurations before
+parsing and listing. The resolved yaml is validated as well.
+
+The component is associated with the `postprocess_on` key. If this key is missing
+or set as True, the component will be post-processed and will be listed. If the key
+is set to false, it will not be listed with the subtool.
```
Ciheim: Should 'false' be capital here?

fre/list_/list_experiments_script.py
```diff
@@ -1,6 +1,11 @@
 """
-Script combines the model yaml with experiment, platform, and target 
-information to list experiment information.
+List_experiments_script provides a method to list experiments
+found in the `model.yaml`. The yaml is resolved in case experiment
+names use fre_properties.
+
```
mlee03: Module `list_experiments_script` contains methods to query the resolved `model.yaml` for `experiment` configurations.  Because the experiment yaml (?) can be interdependent on other configuration yaml files, the methods contained here only work with resolved/combined model.yaml...?

fre/list_/list_experiments_script.py
```diff
@@ -18,9 +23,10 @@
 
 def list_experiments_subtool(yamlfile: str):
     """
-    List the post-processing experiments available
+    List_experiments_subtool loads and parses the model yaml configuration
```
mlee03: `List_experiments_subtool` list the experiment name in `yamlfile`. 
(I'm thinking the loading and parsing the yaml is implied if it needs to list the experiments and does not need to be explicitly stated)

fre/list_/list_platforms_script.py
```diff
@@ -1,6 +1,9 @@
 """
-Script combines the model yaml with the compile and platforms yaml.
-Merged yaml dictionary is parsed to list experiment information.
+List_platforms_script provides a method to list platforms defined in
```
mlee03: Module `list_patforms_script` contains methods to query the `platforms.yaml` configuration file.

Does this also require a resolved yaml?

fre/list_/list_platforms_script.py
```diff
@@ -12,9 +15,12 @@
 
 def list_platforms_subtool(yamlfile: str):
     """
-    List the platforms available
+    List_platforms_subtool uses the "fre yamltools combine-yamls"
```
mlee03: List_patforms_subtool lists the platform name? in platforms.yaml.

fre/list_/list_experiments_script.py
ceblanton: @mlee03 and @singhd789 we definitely need to bring more order to the yaml information process for next year. I think we're almost there, really.

Listing the experiments requires the `unresolved/uncombined` yaml because the combining-yaml process strips out any experiments not specified on the command line (`fre xxx yyy -e expname`), but most information retrieval generally should simply accept the `resolved` python dictionary and then pull out what it needs.

fre/list_/list_platforms_script.py
singhd789: It might yeah, in the case where a reusable variable used in the platforms.yaml is defined in the model yaml, but I guess usually the platform name doesn't reference anything (so maybe this can change...)

fre/list_/list_experiments_script.py
singhd789: Ooook, yeah I wasn't entirely sure either but I added it for explicitness/for the chatbot to have more info. Do you think it's too much and maybe not needed?

fre/list_/list_experiments_script.py
```diff
@@ -1,6 +1,11 @@
 """
-Script combines the model yaml with experiment, platform, and target 
-information to list experiment information.
+List_experiments_script provides a method to list experiments
```
laurenchilutti: I would explicitly say the name of the method. e.g. " List_experiments_script provides **the function list_experiments_subtool which is** a method to ..."

fre/list_/list_platforms_script.py
laurenchilutti: I would explicitly say the name of the method. e.g. " List_platforms_script provides **the function list_platforms_subtool which is** a method to ..."

fre/list_/list_experiments_script.py
laurenchilutti: I agree with @mlee03 that this could be refined to better express what you intent here.  I am chiming in to express that `list_platforms_script.py` and `list_pp_components_script.py` module docstrings should also share a similar explanation/wording about the combined yaml.

fre/list_/list_pp_components_script.py
```diff
@@ -1,6 +1,12 @@
 """
-Script combined model yaml with the settings and post-processing yamls.
-Merged yaml dictionary is parsed to list components that will be post-processed
+List_ppcomps_subtool provides a method to list components to be post-processed,
```
laurenchilutti: I would explicitly say the name of the method. e.g. "List_ppcomps_subtool provides the function list_ppcomps_subtool which is a method to ..."

fre/list_/list_experiments_script.py
singhd789: I thiiiiiink I was doing too much. I removed the bits about even combining the yamls. I'll add it back in if you think it was helpful but the main point of the list subtools is to list the information, so explaining that the yamls could be combined and resolved might be extra.

fre/list_/list_platforms_script.py
singhd789: simplified!

